PR #26, opened 2016-12-21
fix capitalization issues in some of the newer updates

hlarsen:
a few of the updates from this morning have some capitalization issues that are causing the import script to die


killerwife:
will fix it tomorrow after I deal with my exam

Issue #25, opened 2016-12-21
error importing db


hlarsen:
error importing the initial db - it looks like some of the values in game_event are set to 0000-00-00 00:00:00, then when the script tries to add the new column on line 369, it complains because the default for the columns are set to 1970xxxx and the 0000-00-00 is invalid.

happy to help/test if needed.

~~~~
    Appending Updates/998_Update_schema.sql
ERROR 1067 (42000) at line 369: Invalid default value for 'start_time'
ERROR: cannot apply Updates/998_Update_schema.sql
~~~~

killerwife:
I just pushed some changes 10 minutes ago, and I just ran installscript, and it worked perfectly for me.

hlarsen:
I'll test with the new changes when I can.

Just to be clear, should we import all of the 'base' SQL files from the
mangos-tbc repo, then run the tbc-db install script?

killerwife:
just run install script, I supply complete package when adding core changes

hlarsen:
ok, here's the deal:

./Full_DB/TBCDB_1.5.0_cmangos-tbc.sql creates the game_event table with defaults for start_time and end_time of '0000-00-00 00:00:00', then populates the table with some rows, a few of which include '0000-00-00 00:00:00' as start/end times. 

sql_mode NO_ZERO_DATE doesn't like timestamp fields containing all zeroes; this is apparently what is causing the 'invalid default value' error when 998_Update_schema.sql tries to alter the table to add a new column. it also throws an 'incorrect datetime value' when trying to update the rows by using 'where start_date = '0000-00-00 00:00:00', which 998_Update_schema.sql tries to do near the end. we can, however, update the zero date rows by entry.

removing NO_ZERO_DATE from sql_mode fixes this, but it appears it is on by default. obviously a few different ways to fix this, your call - let me know if i can be of any help.

killerwife:
Sheesh, old @cyberium was right, that strict mode is hella annoying. It means I need to apply query before alter table. Can you test that?

PR #28, opened 2016-12-22
fix ordering of game_event table updates, add missing backticks

hlarsen:
this should take care of the import issue; the import script finishes running now. i set the events with blank times to the new default since i wasn't sure of the dates.

fixes cmangos/tbc-db#25


killerwife:
Nice, my wild guess was right :D I suggest you notify other DB devs and also tell @cyberium to shuffle it around in the core file. (he will see this :D)

killerwife:
@hlarsen I think you moved it too far north.

PR #29, opened 2016-12-27
fix capitalization issues

hlarsen:
a few more fixes for table names

### DIFF
--- a/Updates/692_Fix_Incorrect_Plan_Drop.sql
+++ b/Updates/692_Fix_Incorrect_Plan_Drop.sql
@@ -1,3 +1,3 @@
 -- Plans: Ragesteel Helm Should Only Drop From Anger Guards
-DELETE FROM creature_Loot_template WHERE item=23612 and entry IN (21060,21061);
+DELETE FROM creature_loot_template WHERE item=23612 and entry IN (21060,21061);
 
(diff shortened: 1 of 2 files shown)


killerwife:
Is this all changes needed to make it run in strict mode?

ulfgebhardt:
i get the same errors and its cuz of the storage method. Linux stores files case-sensitive. Therefore Tablenames are case-sensitive.

More here:
http://dev.mysql.com/doc/refman/5.7/en/identifier-case-sensitivity.html

killerwife:
I should update my server I think.....

ulfgebhardt:
Its a linux thingy, if u use windows, u wont get that error for tables, triggers and databases - if u use linux this should always be the case. Could u also update the Loot References? Pretty please <3

killerwife:
Yes its on my TODO list, but as I neglected vengeance for the past 3 weeks due to exporting these changes, I have to do a lot of cleanup atm. Will do it whenever I have time.

ulfgebhardt:
vengeance is ur server? Nice ;-)

killerwife:
Yes, and these are all vengeance changes, all the 600+ commits I added. (ofc some being backports)

killerwife:
Ty, merging.

PR #35, opened 2017-01-22
Watchkeeper Gargolmar

AnonXS:
Open Issues from https://jira.vengeancewow.com/browse/TBC-2084

Mortal Wound should stack up to 10 not 8
http://wowwiki.wikia.com/wiki/Watchkeeper_Gargolmar?oldid=1479057

Adds should respawn on evade

---

https://github.com/cmangos/mangos-tbc/blob/master/src/scriptdev2/scripts/outland/hellfire_citadel/hellfire_ramparts/boss_watchkeeper_gargolmar.cpp#L38

### DIFF
--- a/Updates/842_AnonXS_WatchkeeperGargolmar.sql
+++ b/Updates/842_AnonXS_WatchkeeperGargolmar.sql
@@ -1,0 +1,3 @@
+UPDATE `spell_template` SET `StackAmount` = 10 WHERE `id` IN (30641, 36814);
+
+UPDATE `creature_linking` SET `flag` = `flag`|4 WHERE `master_guid` = 62083;


killerwife:
Hmmm, are you sure that the 10 stacks thing is not related to patch 2.0.3? Because progression wise, public tbc-db is patch 2.4.3.

AnonXS:
Latest http://wowwiki.wikia.com/wiki/Watchkeeper_Gargolmar: stacks up to 10 times 

http://www.wow-pro.com/dungeon_and_raid_guides/5862_hellfire_citadel_ramparts

Guide published  2008-02-22

https://web.archive.org/web/20100327112713/http://wow.gamona.de/das-spiel/instanzen-und-raids/burning-crusade/das-bollwerk-heroic

Guide published maybe even later with 100% Heal Reduce.

https://web.archive.org/web/20100127135927/http://www.hordeguides.de/index.php?ib=17306

15.06.2008

killerwife:
Thanks, also for the future I would like to point out that spell_template queries are stored in core. But for the meantime I will always take care of that myself. Will process this soon-ish, when I get the data for closing honor hold PR.

killerwife:
The spell_template line will be merged at a later date when I export a lot of spell_template work from vengeance. ETA within next month.

PR #39, opened 2017-01-26
Nagrand - Bach'lor movement and linking

Saltgurka:
He's supposed to stop at a few points and do random movement. 

Unfortunately I can't do that with SCRIPT_COMMAND_MOVEMENT, because when I set him back to waypoint movement he starts over from waypoint 1. 
I added some closeby waypoints to simulate random movement instead.


MantisLord:
It seems you're right about the lack of support to resume waypoint movement at same point after switching to random movement.
Creature always goes to first point upon switching back to waypoint movement with SCRIPT_COMMAND_MOVEMENT.
I propose a simple core change to allow us to set current waypoint using datalong in dbscripts:

```
--- a/src/game/ScriptMgr.cpp
+++ b/src/game/ScriptMgr.cpp
@@ -1618,6 +1618,10 @@ bool ScriptAction::HandleScriptStep()
                     break;
                 case WAYPOINT_MOTION_TYPE:
                     ((Creature*)pSource)->GetMotionMaster()->MoveWaypoint();
+
+                    if (m_script->movement.wanderDistance)
+                        ((Creature*)pSource)->GetMotionMaster()->SetNextWaypoint(m_script->movement.wanderDistance);
+
                     break;
             }
```
With this change, we could get movement pattern you are talking about using dbscripts like this:
```
(6552801,0,20,1,20,0,0,8,0,0,0,0,0,0,0,0,'Bachlor - Random movement around current position'),
(6552801,30,20,2,4,0,0,0,0,0,0,0,0,0,0,0,'Bachlor - Waypoint movement starting at point 4'), 
```

killerwife:
The better design would be to pause waypoint movement and initiate temporary random movement. This is not the first time I am seeing a limitation in our random movement. Problem is that there is no simple answer for this issue. It is quite apparent that we have to redo random movement and make it more configurable, for the sake of better scripts. I will keep this on my radar, but i cant promise anything.

PR #42, opened 2017-01-29
Molten Core Rework

Saltgurka:
- Full rework of Molten Core including fixes to linking, movement, respawns, pooling, creature stats, immunities etc.

Credits and sources inside the file.

Also ported to wotlk-db: https://github.com/cmangos/wotlk-db/pull/66
and classic-db: https://github.com/cmangos/classic-db/pull/33


killerwife:
Did u make a rough run through the instance? Just want to make sure we can summon ragnaros after dousing all the runes.

Saltgurka:
Yes that does work for the most part. It seems to me like if no player is near Majordomo's area when you click the last rune he won't spawn at all. Not sure if that was the actual cause for it not working, but it sure seemed like it. 
Clicking Sulfuron Harbinger's rune last it worked every time. Going back and clicking for example Gehenna's rune last, and he didn't spawn.

killerwife:
Ok, I am sure I can find what causes that, whenever I get time to review/merge this. Good job tho.

PR #44, opened 2017-02-11
Full Rework Battleground Holidays Emissaries

AnonXS:
which are especially important for vanilla and tbc.
Some gameobjects are still missing, didnt bother as npcs are most important.


Tobschinski:
And where does this data come from?

AnonXS:
tdb

PR #49, opened 2017-02-12
Blade's Edge Mountains - Mature Cavern Crawler

Tobschinski:


Updates/SpinOnTop/xxx_mature_cavern_crawler.sql
```diff
@@ -0,0 +1,83 @@
+
+SET @GUID := 78185;
```
Tobschinski: ^ check for free guid

PR #51, opened 2017-02-14
Update 044_UDB_BACKPORT_PART1.sql

ekg98:
Correcting to lower case.  Ty for your guys hard work.  ekg98


killerwife:
Only case issue?

ekg98:
Yes installed fine after change of case.  From Creature to creature.

PR #55, opened 2017-02-16
Monsteritem - Template

Tobschinski:
lots of known monsteritems (only vanilla, items added in tbc are missing)


Tobschinski:
Will add this when I remove creature_equip_template_raw.

PR #58, opened 2017-02-19
TBC - Equipment

Tobschinski:
adds missing equipment to about 1600 creature_templates
fixes lots of incorrectly assigned equipment in the Outlands


Tobschinski:
Existing data was moved, not yet removed.

There's about 130 creatures still using old entries in either creature_equip_template_raw or creature_equip_template. They are more complicated (e.g. choose out of 4  random equipment sets)

Another thing that still needs to be done is to check the `creature` for 455 creatures that override their template equipment.

killerwife:
Random equipment sets? That thing exists?

Tobschinski:
I guess so, I found this item ...
('30702', '0', '0', 'Monster - EQUIPS RANDOM WEAPON VIA ACTION TRIGGER', '7040', '13', '0', '3'),

killerwife:
Interesting :) Well, should you want me to implement this, you need to make a rough draft of what it should be capable of.

AnonXS:
Same with http://www.wowhead.com/npc=18423/chowar-the-pillager.
https://youtu.be/sbUTfQS8yDU?t=2m40s , https://youtu.be/GEOy2JkUWNU?t=1m37s

Tobschinski:
There's a few more, yes

cooler-SAI:
very good thing. Thanks)))

cooler-SAI:
can be merged ?

PR #63, opened 2017-02-26
Hero of the Maghar Thrall Event

Saltgurka:
Example vids for reference: https://www.youtube.com/watch?v=cuQCL7z9fMc , https://www.youtube.com/watch?v=VBiapet3a9g

This is the event that starts after a player turns in [Hero of the Mag'har](http://db.hellfire-tbc.com/?quest=10212)

Some notes:

- ~~Equipment is currently missing for some of the event NPCs.~~ Solved with https://github.com/cmangos/tbc-db/pull/58

- ~~https://github.com/cmangos/mangos-tbc/pull/174 needs to be implemented for this event to work. (For multiple creature_movement_template pathIds and for TRIGGERED_IGNORE_UNSELECTABLE_FLAG)~~

- Kor'kron Warrior/Honor Guards currently have their own separate paths. Once creature formations are implemented they should be divided into two groups with Honor guards in the middle: 
image

- Garadar Wolf Riders should possibly stop and greet Thrall if they run into him inside Garadar. I've been unable to confirm or witness this myself, and I'm not sure how that would be possible to implement anyways so it's left out for now.

- ~~Sniffed flag for Mannoroth is 33555200. However UNIT_FLAG_NOT_SELECTABLE makes the  animation from Thrall's Throw Doomhammer spell be invisible, so I had to remove that flag.~~

- The buff casted by Garrosh's Buff Bots(34410) at the end of the event is stacking, so if you're standing close to more than one buff bot you will get duplicate buffs.

- ~~For some reason the buff casted by Garrosh's Buff Bots(34410) does not hit Thrall and Garrosh. Factions seem to be correct, so I'm not sure why it's not working.~~

- The maximum search distance for SCRIPT_COMMAND_TERMINATE_SCRIPT is 500y. Thrall spawns further away than that from the quest turn in, so it is possible for players to sometimes start the event again even if Thrall is already spawned.

- The text emote on Thrall spawn ("War drums echo in the distance.") should be visible all the way to Garadar, but we have no zone wide text emote type.

- The music should change to id 6146 when the event starts and stay that way until Thrall reaches Garrosh. Most likely this should be triggered by the NPC named Thrall's Hero Music (19597). However it is currently not possible to change the music zonewide, so for now Thrall is changing the music for those near him only.


killerwife:
"Sniffed flag for Mannoroth is 33555200. However UNIT_FLAG_NOT_SELECTABLE makes the animation from Thrall's Throw Doomhammer spell be invisible, so I had to remove that flag."
This is already solved in EAI, I enable passing of special triggered flags. I will add support to DBScript now that we have datalong3 in my PR. Its called TRIGGERED_IGNORE_UNSELECTABLE_FLAG in C++

Zone wide music, are you sure? Not just very big distance like in Darkmoon Faire or Lament of the Highborne, or Brewfest?

Also zone wide emote? I have reason to believe it should be distance based, not zone based. Would be nice if you could verify that u can see the emote from furthest corner of nagrand. (Its easy to implement, but we encountered something similar in karazhan with animal boss so I need data :) )

Does thrall and garrosh by any chance have any unattackable flag? That would explain why 34410 cant hit him.

34410 Stacking issue has been given to @WarlockBugs and he will deal with it once he is again playing with stacking.

Greeting in EAI is done using EVENT_T_OOC_LOS but only works for player at the moment. Should the eventuality arise and be needed for two creatures let me know.

Saltgurka:
> Zone wide music, are you sure? Not just very big distance like in Darkmoon Faire or Lament of the Highborne, or Brewfest?
> 
> Also zone wide emote? I have reason to believe it should be distance based, not zone based. Would be nice if you could verify that u can see the emote from furthest corner of nagrand. (Its easy to implement, but we encountered something similar in karazhan with animal boss so I need data :) )

It's possible that it is just a very large range for both the music and text emote.  To test it I need help from someone though. One need to turn in the quest and the other stay in a far away corner.


> Does thrall and garrosh by any chance have any unattackable flag? That would explain why 34410 cant hit him.

Yes, they both have UNIT_FLAG_OOC_NOT_ATTACKABLE. I removed it and it works. Would TRIGGERED_IGNORE_UNSELECTABLE_FLAG  work for this spell too or is there another workaround?

killerwife:
No, but its something to consider to be added with it :) Cos its not the first time I am seeing it. Blizzard has a different serverside setting it seems. I.e. something like "cant be hit by player, but can be hit by NPC stuff" and we dont have such capabilities at the moment.

Saltgurka:
For Mannoroth TRIGGERED_IGNORE_UNSELECTABLE_FLAG worked like a charm. 

TRIGGERED_IGNORE_UNATTACKABLE_FLAG did not work for 34410. Neither Garrosh or Thrall got the buff. Strange thing is that even when I make them cast it themselves, either through dbscript or ingame with .cast back 34410.

I guess UNIT_FLAG_OOC_NOT_ATTACKABLE is different from UNIT_FLAG_NOT_ATTACKABLE_1?

Saltgurka:
So an update on the music. The range is all the way to where I am on the map below:
image


Going further away than that, and it switches to normal nagrand music. I haven't looked into the emote and yell (I was only lucky to catch someone else start the event on retail), but I would assume their range is the same. 

Another question now is if it is solely a range check, cause in that case it will affect players in shattrath and zangarmarsh as well (since the origin is at the spawnpoint of thrall). Maybe there is a check for zone as well, so only players in Nagrand hears it? Needs some research...

PR #66, opened 2017-02-28
[ACID] Improve OOC Scripts for 18292,18293,18296,19140

Saltgurka:
* Previously all Bleeding Hollow Refugees/Sunspring Post Refugees/Sunspring Post Orphans were doing these texts and emotes. Now only the Refugees/Orphans standing around the Pitfigher does emotes and texts.
* Also added one missing text (reused -317) and added some phase logic to allow all 6 texts to be randomized

It works like this:
* The pitfighter throws AI Event 5 in a 5 yard radius. 
* The Orphans/Refugees that receive the event sets phase to 1.
* In phase 1 Refugees will cast random emotes every 1500-2500ms
* In phase 1 Refugees and Orphans will set phase to 2 or 3 randomly every 5000-250000ms
* In phase 2 Refugees and Orphans will do a random text and set phase 1
* In phase 3 Refugees and Orphans will do a random text and set phase 1

Edit: Forgot entry 18296 in the commit message


Saltgurka:
I noticed some additional issues and more NPCs that this behaviour should apply to. Will open a new PR instead

PR #67, opened 2017-02-28
Kurenai/Mag'har Pitfighter (+ their spectators) script improvements

Saltgurka:
* Improvements to NPCs 18292,18293,18296,18488,19140,19141,19149.
* Previously ALL Bleeding Hollow Refugees/Sunspring Post Refugees/Sunspring Post Orphans were doing these texts and emotes. Now only the Refugees/Orphans standing around the Pitfigher does emotes and texts.
* Telaari Watchers and Telaari Citizens were missing texts and emotes completely.
* Added a missing spell to Kurenai/Mag'har Pitfighter.
* Added one missing text (reused -317) and added some phase logic to allow all 6 texts to be randomized.
* Added separate texts for Telaari NPCs.
* Updated timers for both pitfighters and spectators.
* Nagrand Target Dummy (19139) script removed from EventAI, now handled in dbscripts_on_spell.
* Nagrand Target Dummy will now do emotes on spellhit instead of OOC Timer.
* Removed one incorrect spawn.

It works like this:
* The pitfighter throws AI Event 5.
* The Spectator NPCs that receive the event sets phase to 1.
* In phase 1 Spectators will cast random emotes every 1500-2500ms
* In phase 1 Spectators will set phase to 2 or 3 randomly every 5000-250000ms
* In phase 2 Spectators will do a random text and set phase 1
* In phase 3 Spectators will do a random text and set phase 1

Note that we currently do not have any Telaari Watchers spectating in Telaar, but the proper implementation is that they should be pooled with the Telaari Citizen. I will add that in the future (if no one else beats me to it), but for now there's no harm to implement their scripts before spawns are added.


Saltgurka:
Error:
> 2017-02-28 19:40:23 Event 1829203 Action 2 attempts to set phase3 >= 32. Phase mask cannot be used past phase 31.

That's referring to `('1829203','18292','1','13','100','1','5000','250000','5000','250000','0','0','0','0','30','2','3','-1','0','0','0','0','Bleeding Hollow Refugee - Set Phase 2/3 randomly (Phase 1)'),`

But documentation states that -1 should be used if you're only randomizing between 2 phases. 
https://github.com/cmangos/mangos-wotlk/blob/master/doc/EventAI.txt#L746
> NOTE: Use -1 to specify that if this param is picked to do nothing.

Am I not understanding that correctly?

Saltgurka:
The additional phases that was added is required for this PR: https://github.com/cmangos/tbc-db/pull/68

Just pull in the Orphans waypoints and dbscript in order to test the new phases: https://github.com/cmangos/tbc-db/pull/68/files#diff-bc7736a3d286688e9ab28027776c7176R316

PR #70, opened 2017-03-09
Skettis - v2.1+

Tobschinski:
First part (of many) that will upgrade Skettis to its version used after 2.1


Rushor:
all those changes can be merged to the cmangos wotlk db too? :)

Tobschinski:
Can't tell, I would have to check first, maybe @Grz3s can ...

PR #88, opened 2017-05-29
Bug fix to 2 quest and 1 creature

MrSticky:
Quest 10641 - https://jira.vengeancewow.com/browse/TBC-1054
Quest 10078 - https://jira.vengeancewow.com/browse/TBC-2544
Creature 18694 - https://jira.vengeancewow.com/browse/TBC-2542


Saltgurka:
Thank you for your contribution!

Quest 10641 pushed here: https://github.com/cmangos/tbc-db/commit/e48f47d72369c82392f26406788968f66ec517e9
and the creature path fixed here: https://github.com/cmangos/tbc-db/commit/c8ed5859f37e231aaa3ecf9cb75645cb8278a58f

I'm not so sure about this part though: `update quest_template set ReqItemId3=0 where entry=10078;`
The side effect of that is that the torch is not consumed upon completing the quest, which it should be as far as I know.

Saltgurka:
I'm closing this for now as I don't think the fix for q10078 is correct, since the item would not be consumed on quest completion. It's possible that this is a core issue and that the objective should only be displayed as completed if both the item and the creature requirements are met.

PR #94, opened 2017-07-15
[ACID] Bat Rider Guard 15242

AnonXS:
http://www.wowhead.com/npc=15242/bat-rider-guard
http://www.wowhead.com/spell=38066/unstable-concoction#comments

- should attack after players get marked by Guard's Mark which currently does not work.

By raum (592 –  4·7)  on 2007/10/16 (Patch 2.2.3)
This is the effect you receive when enemy flyers attack you after receiving Guard's Mark. It will knock you off your mount, instantly killing you if you are too high (from hitting the ground).


killerwife:
Updated fork? :D

AnonXS:
cant push a pr to tbc-db directly is what it tells me. . .

Saltgurka:
Are you sure it should only be cast on initial aggro and not on a combat timer?

AnonXS:
I am not sure, but that's how i found this script. comments suggest that it can be cast repeatedly, but maybe the spell was changed after tbc, which seems likely from these comments. I would second a repeated combat timer, but have no evidence for it.

By loktar09 (2,652 –  2·28)  on 2010/05/10 (Patch 3.3.3) |
has the ability been nerfed ive been hit 5 times in a row on a lvl 61 DK and didn't die or dismount
can anyone confirm if they have?

Forkmi on 2011/01/04 (Patch 4.0.3)
I was hit like 7 times in a row as level 60 night elf rogue, and all it did was lose me 20% health. Didn't dismount me or anything.

Saltgurka:
Merged in https://github.com/cmangos/tbc-db/commit/75d4efb68092548d51b9b445c577cac5c81d2168

PR #95, opened 2017-07-17
[ACID] Improve 16925 and 18706

Darexon:
Bonechewer Raider (16925) now has chat texts and unmounts when aggro'd
Bonechewer Riding Wolf (18706) now has scripted abilities and despawns when exiting combat
Bonechewer Scavenger (18952) now has more chat texts and properly working phases
Fixes cmangos/issues/issues/1424

### DIFF
--- a/ACID/acid_tbc.sql
+++ b/ACID/acid_tbc.sql
@@ -18156,7 +18156,8 @@ INSERT INTO `creature_ai_scripts` (`id`,`creature_id`,`event_type`,`event_invers
 ('1691208','16912','2','3','100','0','15','0','0','0','25','0','0','0','1','-47','0','0','0','0','0','0','Mag\'har Hunter - Flee at 15% HP (Phase 2)'),
 ('1691209','16912','7','0','100','0','0','0','0','0','22','1','0','0','40','1','0','0','0','0','0','0','Mag\'har Hunter - Set Phase 1 and Set Melee Weapon Model on Evade'),
 -- Bonechewer Raider
-('1692501','16925','4','0','100','0','0','0','0','0','11','32723','0','0','0','0','0','0','0','0','0','0','Bonechewer Raider - Summon Bonechewer Riding Wolf on Aggro'),
+('1692501','16925','4','0','100','0','0','0','0','0','43','0','0','0','11','32723','0','0','0','0','0','0','Bonechewer Raider - Dismount and Cast Summon Bonechewer Riding Wolf on Aggro'),
+('1692502','16925','4','0','20','32','0','0','0','0','1','-765','-766','0','1','-767','-768','0','0','0','0','0','Bonechewer Raider - Say on Aggro'),
 -- Stonescythe Alpha
 ('1692901','16929','4','0','100','0','0','0','0','0','22','1','0','0','0','0','0','0','0','0','0','0','Stonescythe Alpha - Set Phase 1 on Aggro'),
 ('1692902','16929','9','5','100','1','0','5','5000','9000','11','33911','1','0','0','0','0','0','0','0','0','0','Stonescythe Alpha - Cast Tear Armor (Phase 1)'),
@@ -18271,6 +18272,9 @@ INSERT INTO `creature_ai_scripts` (`id`,`creature_id`,`event_type`,`event_invers
 -- Vorakem Doomspeaker
 ('1867901','18679','9','0','100','1','0','5','6000','9000','11','9080','1','0','0','0','0','0','0','0','0','0','Vorakem Doomspeaker - Cast Hamstring'),
 ('1867902','18679','2','0','100','0','20','0','0','0','11','8599','0','1','0','0','0','0','0','0','0','0','Vorakem Doomspeaker - Cast Enrage at 20% HP'),
+-- Bonechewer Riding Wolf
+('1870601','18706','9','0','100','1','0','5','10500','12500','11','16460','1','0','0','0','0','0','0','0','0','0','Bonechewer Riding Wolf - Cast Festering Bite'),
+('1870602','18706','0','0','100','1','8000','8000','15000','15000','11','3149','0','0','0','0','0','0','0','0','0','0','Bonechewer Riding Wolf - Cast Furious Howl'),
 -- Kuma
 ('1878501','18785','4','0','100','0','0','0','0','0','12','9527','1','30000','12','9527','1','30000','1','-549','0','0','Kuma - Summon Enraged Hippogryphs on Aggro'),
 -- Vlagga Freyfeather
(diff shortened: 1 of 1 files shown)


AnonXS:
Maybe add Texts from http://www.wowhead.com/npc=16925/bonechewer-raider too if they are legit: 

```
('1692501','16925','4','0','100','0','0','0','0','0','43','0','0','0','11','32723','0','0','1','-765','-766','-767','Bonechewer Raider - Dismount and Cast Summon Bonechewer Riding Wolf and Say on Aggro'), 
.
.
.
('-766','For Spinebreaker!','0','0','0','16925','0'),
('-767','Protect the Warchief at all costs!','0','0','0','16925','0'),
```

Darexon:
The only one out of those that I have not seen is "Protect the Warchief at all costs!" but I have seen "Your bones will be devoured!". Bonechewer Raiders and Bonechewer Scavengers share the same texts and the texts seem to be rare. Judging by the linked videos I think it is a 5% chance.
https://www.youtube.com/watch?v=qDwlMUt4tf8
https://www.youtube.com/watch?v=IygZHwEWaiA
https://www.youtube.com/watch?v=Mg2frECdjmM

AnonXS:
Yes, text spam is not blizzlike. :)

Saltgurka:
Script 1895201 definitely seems incorrect, yes. The text should have its own event.

Regarding this pr, it is definitely an improvement over the old script. However the way I think it should work after playing around with these npcs on retail is that the wolf despawn and the raider mount up should always happen at the same time. The pet should despawn when the raider evades, and the raider should instantly mount up. The pet itself does not seem to have a max evade/leash distance, the raider always evades first. 

Unfortunately as far as I know we have no way to tell a creature to despawn its own pet in EventAI and simply throwing an AI Event could potentially despawn another raider's pet. We might want to add a small SD2 script to achieve that.

Another issue I noticed, which is probably a core issue, is that if you kill the pet but not the raider, after 5mins the pet will respawn and walk beside the raider.

ACID/acid_tbc.sql
```diff
@@ -18287,12 +18292,13 @@ INSERT INTO `creature_ai_scripts` (`id`,`creature_id`,`event_type`,`event_invers
...
 -- Bonechewer Scavenger
-('1895201','18952','4','0','20','0','0','0','0','0','1','-765','0','0','22','1','0','0','0','0','0','0','Bonechewer Scavenger - Say and Set Phase 1 on Aggro'),
-('1895202','18952','0','5','100','0','1000','3000','0','0','11','13398','1','0','0','0','0','0','0','0','0','0','Bonechewer Scavenger - Cast Throw Wrench (Phase 1)'),
-('1895203','18952','9','5','100','1','0','5','6000','9000','11','34113','1','0','0','0','0','0','0','0','0','0','Bonechewer Scavenger - Cast Bonechewer Bite (Phase 1)'),
-('1895204','18952','2','0','100','0','15','0','0','0','22','1','0','0','0','0','0','0','0','0','0','0','Bonechewer Scavenger - Set Phase 2 at 15% HP'),
-('1895205','18952','2','3','100','0','15','0','0','0','25','0','0','0','1','-47','0','0','0','0','0','0','Bonechewer Scavenger - Flee at 15% HP (Phase 2)'),
-('1895206','18952','7','0','100','0','0','0','0','0','22','1','0','0','0','0','0','0','0','0','0','0','Bonechewer Scavenger - Set Phase 1 on Evade'),
+('1895201','18952','4','0','20','0','0','0','0','0','22','-1','0','0','0','0','0','0','0','0','0','0','Bonechewer Scavenger - Set Phase 1 on Aggro'),
```
Saltgurka: Chance to set phase 1 should be 100%

Darexon: Working on it. It also sets the wrong phase and the inverse phase masks are wrong.

Saltgurka:
Closed the wrong pr, my bad

Darexon:
I think my work is done i'm just not sure about: The chance of saying something being 10% (it was 20% before on the Bonechewer Scavenger), If other Bonechewer Orcs have these texts (in the linked videos I've only saw Bonechewer Raider and Bonechewer Scavenger use them), Bonechewer Scavenger only throwing wrench once in combat and doing it at the start of combat.

Saltgurka:
`Protect the Warchief at all costs!` is legit as well
image
image

I think you're right that only Scavengers and Raiders say the texts.

Darexon:
So how would I include those 4 texts on one NPC? If I create a new event and lower chances there's a possibility of the NPC saying 2 texts at once.

Saltgurka:
You can use [EFLAG_RANDOM_ACTION](https://github.com/cmangos/mangos-wotlk/blob/master/doc/EventAI.txt#L1007) to randomize between actions. 
It would look something like this:
```sql
('1692502','16925','4','0','10','32','0','0','0','0','1','-765','-766','0','1','-767','-768','0','0','0','0','0','Bonechewer Raider - Say on Aggro'),
```
This will randomize between the two actions and the actions themselves will randomize between the two texts. That way only one text will be said

Darexon:
I've removed the wolf evade in EventAI so the despawning could instead be done in C++. As for the code you suggested I made it match the style of the rest of the file and build with these changes
```
/*######
## npc_bonechewer_raider
######*/

struct npc_bonechewer_raiderAI : public ScriptedAI
{
    npc_bonechewer_raiderAI(Creature* pCreature) : ScriptedAI(pCreature)
    {
        Reset();
    }

    void Reset() override {}

    void EnterEvadeMode() override
    {
        m_creature->GetPet()->ForcedDespawn();
        m_creature->Mount(17408);
    }

    void UpdateAI(const uint32 uiDiff) override
    {
        DoMeleeAttackIfReady();
    }
};

CreatureAI* GetAI_npc_bonechewer_raider(Creature* pCreature)
{
    return new npc_bonechewer_raiderAI(pCreature);
}
```
When I tested it in game the wolf never despawned.

AnonXS:
I would have wondered if GetPet worked for npcs :D

Saltgurka:
On second thought, I'm not sure we can have both EventAI and an SD2 script for the same creature. 

I'll ask kw what would be the easiest solution here and I'll process the PR as soon as possible.

cala:
No, you can't have both: there is only one AI active per creature. So, if you cannot do what you want with EventAI, SD2 is probably what you need.

MantisLord:
Did not mean to auto close this.

Saltgurka:
This PR (https://github.com/cmangos/mangos-wotlk/pull/228) will add the EventAI action ACTION_T_DESPAWN_GUARDIANS, which is exactly what we need for the Bonechewer Raider.

I think it's better to wait for that to be merged first than to add an SD2 script that will be obsolete very soon. I will get back to this PR as soon as core part is merged :)

AnonXS:
Same Issue with: http://www.wowhead.com/npc=19640/bleeding-hollow-riding-worg, http://www.wowhead.com/npc=19424/bleeding-hollow-tormentor, They need similar AI, maybe include when we have core support for the underlying issue.

Saltgurka:
@AnonXS The Bleeding Hollow Tormentor is a harder one because not all of them are mounted to begin with, so we would need to do some check to only summon wolves if the master was mounted to begin with

PR #112, opened 2017-08-11
[NPC] Damageschool Update

AnonXS:
There are possibly alot more "Elemental" Npcs which should do elemental damage.

I dont have any sources for the npcs in the second file maybe someone can confirm these.


Saltgurka:
I was able to confirm most of the uncornfirmed ones in the second file.

I only found these two to be incorrect:
```sql
-- Unconfirmed ingame patch 7.3 (https://i.imgur.com/XkVZdpj.png)
UPDATE `creature_template` SET `damageschool` = 3 WHERE `entry` = 21253;

-- Unconfirmed ingame patch 7.3 (https://i.imgur.com/U7O9X5C.png)
UPDATE `creature_template` SET `damageschool` = 4 WHERE `entry` = 21260;
```
Pushed as a combined file here: https://github.com/cmangos/tbc-db/commit/5b3e8cbc05196d32d8bdd7df9611a50547fce7d1

PR #114, opened 2017-08-11
[ACID] Ravenous Flayer 21196

AnonXS:
Spell http://www.wowhead.com/spell=37933/consume-flesh seems to currently heal the target (player) for as much as it deals damage, should heal the caster (npc) for as much as it deals damage.

**Drains health from an enemy every 1 sec., transferring it to the caster.**

http://www.wowhead.com/npc=21196/ravenous-flayer#comments:id=194331


killerwife:
Oh, today I added a usecase for arcane missiles I broke that fixes this kind of interaction. Sounds like I need to think up a generic behaviour ^^

PR #133, opened 2017-08-12
Create Fix - Zandalar Items Required Level

dunjeon:
Zandalar Items Required Level should be 60 not 0


AnonXS:
Not a good idea as Replace works like this: https://dev.mysql.com/doc/refman/5.7/en/replace.html.
All other colums in that table will be set to default value for the primary keys you are trying to update.

UPDATE or DELETE & INSERT.

dunjeon:
Absolutely, I will correct it.

Saltgurka:
What is your source for this? From what I can gather these items never had a level requirement, only the quests awarding them does.

Did you find information suggesting otherwise?

dunjeon:
Hey mate,

No new information was mined on my wowhead scrape, but items with an item level of ~60 while having a required level of 0 seems to be off...  The gate to get the gear is likely enough to prevent players from getting it (level and raid requirements to even enter the instance to farm the high level mobs for the drops needed to turn in for the gear).  However, it became an issue while putting a script together to gear bots -- countered by disallowing items that are anything beyond bind on equip.

thanks for taking a look

Warlockbugs:
In general there is pathway to get equipment of much higher ilvl and stats than it would be possible to equip at player's level through quest rewards. Most of equipment quest rewards (all of them?) rely on quest level restrictions instead. The design decision behind this is to reward the player for leveling in groups and overextending. This is one of the main techniques used to gear up low level twinks.

PR #135, opened 2017-08-13
Create Fix - SFK Courtyard Door

dunjeon:
Corrects the broken courtyard door script in Shadowfang Keep


Saltgurka:
This seems to already exist in the scriptdev2.sql file (https://github.com/cmangos/mangos-tbc/blob/master/sql/scriptdev2/scriptdev2.sql)

Perhaps you forgot to specify the core path in your DB install script configuration, resulting in that file not being applied?

dunjeon:
I will check a fresh DB and see if thee error is still there.

dunjeon:
This is what I get on a fresh database with updates, acid and sd2 loaded.  Can I get a confirm on this? (perhaps it is unique to me).

sfk_error

dunjeon:
loading sd2 after all updates have been applied corrects the issue -- do it before updates and you should be able to recreate the error.

PR #136, opened 2017-08-14
Create Fix-IncorrectLowDamage

dunjeon:
Mobs should not have a Damage Multiplier of 0 -- 1 is default.


AnonXS:
Make it a .sql file and dont include your select query. :)

dunjeon:
Done, thanks. =)

Saltgurka:
Solved here: https://github.com/cmangos/tbc-db/commit/45893bfaa03f67fba5a4e5dbb0971312dab42c1b
Any template that had DamageMultiplier=0 in tbc-db now has the value from wotlk-db.

Mobs should hurt a bit more now :)

PR #138, opened 2017-08-15
[ACID] Stonescythe Whelp 16927

AnonXS:
http://www.wowhead.com/npc=16927/stonescythe-whelp#comments:id=47831

- Is missing corresponding spell scripting, which should make the unit stop attacking during the duration of the spell 33896


Saltgurka:
@killerwife 
Looks like we need the serverside spells 33897 and 30023 here
image

killerwife:
https://github.com/cmangos/mangos-tbc/commit/cd99d8fe8f5a761933158408896321ae3cba4718

AnonXS:
Creature doesnt start attacking/moving again, when aura fades.
unbenannt

PR #143, opened 2017-08-15
Revert recent Voodoo Slave update

Phatcat:
Shouldn't have been up-ported from classic.
Was changed during tbc to this.


Saltgurka:
Thanks! We'll try to figure out the Gurubashi Warrior soon as well :)

PR #150, opened 2017-09-04
[ACID] 17134 18260 18262

AnonXS:
http://www.wowhead.com/spell=742/pulverize
http://www.wowhead.com/spell=2676/pulverize

http://www.wowhead.com/npc=17134/boulderfist-crusher#abilities
http://www.wowhead.com/npc=18260/boulderfist-invader#abilities
http://www.wowhead.com/npc=18262/unkor-the-ruthless#abilities


Saltgurka:
I rechecked with sniffs and Aura 742 is not used, they cast spell 2676 directly. Change them to use a normal cast event and I'll merge it. Timer is about 1300ms-20000ms for all three mobs

PR #152, opened 2017-09-10
[ACID] Ortor of Murkblood 18204

AnonXS:
http://www.wowhead.com/npc=18204/ortor-of-murkblood

unbenannt

- [ ] Two Totems missbehavior aura


Saltgurka:
Turns out wowhead is missing two of his spells.
He also uses these two: http://www.wowhead.com/spell=32137/corrupted-air-totem#summons
and http://www.wowhead.com/spell=32135/corrupted-healing-stream-totem

The totems currently don't do anything on summon in cmangos. I've never worked on totems before so not sure what is the right way to fix them. creature_template_addon? creature_template_spells?

AnonXS:
It seems to not make any difference, problem is these two spells (32136, 32138) use Id: 35 (SPELL_EFFECT_APPLY_AREA_AURA_PARTY) which makes him summon totems with auras which should have negative effect on himself but dont due to spell not harming summoner.

He is also not able to summon more than one "corrupted"-totem/guardian at a time, maybe because of EffectMiscValueB = 63 / EffectMiscValueB = 82 which could be a guardian slot.

PR #166, opened 2017-09-23
[ACID] Haal'eshi Windwalker 16966

AnonXS:
http://www.wowhead.com/npc=16966/haaleshi-windwalker


Saltgurka:
I readded the dynamic movement from your first commit since I believe that was the correct way to do it

Pushed here: https://github.com/cmangos/tbc-db/commit/2f35399a28397404af207ef7e582711866f76874

PR #176, opened 2017-10-05
[ACID] Cabal Assassin 18636

AnonXS:
http://www.wowhead.com/npc=18636/cabal-assassin

Comments also point out Sap, Cheap Shot, Poisons maybe even a dual wield ability instead of decreased attacktimer (1500).


Phatcat:
billede

Yes should know dual wield.

Saltgurka:
Since we don't know how Dual Wield should work, let's hold off on adding that for now. Once we have the core implementation done we can do a big Dual Wield update to add it to any mob needing it.

PR #179, opened 2017-10-06
[ACID] Sister of Grief 16960

AnonXS:
http://www.wowhead.com/npc=16960/sister-of-grief


Phatcat:
billede
should also know Seduction and Absorb Life according to Bestiary.

Or Whipped Fury according to wowhead; https://web.archive.org/web/20110720041203/http://www.wowhead.com:80/npc=16960#abilities

I'd go with bestiary over wowhead.

Saltgurka:
The only spell they had when I tested them on retail was Lash of Pain. It is possible that the other two spells listed in bestiary are the spells they get when a warlock enslaves them. It is also possible (altough less likely) that those spells were removed at some point in a later patch.

I think at this point we would need a video or a sniff before adding them

PR #190, opened 2017-10-09
Add Aura for Umbrafen Eel 18138

AnonXS:
http://www.wowhead.com/npc=18138/umbrafen-eel


Saltgurka:
I think this should probably be done in ACID with EVENT_T_MISSING_AURA instead since this aura has a duration. When doing it in creature_template_addon the aura will not be reapplied after it runs out until the creature resets or respawns.

AnonXS:
http://www.wowhead.com/npc=22182/lightning-wasp#comments

http://www.wowhead.com/npc=20290/lagoon-eel#comments

PR #192, opened 2017-10-09
[ACID] Murkblood Invader 18238

AnonXS:
http://www.wowhead.com/npc=18238/murkblood-invader

https://www.youtube.com/watch?v=3NYzpbkVKp0

Works fine on first evade, after that only master_guid = 65509 mounts on evade, 65510+65511 wont.

unbenannt

- [ ] linked guids might evade too soon making it unneeded for them to mount in AI


Saltgurka:
I've run into this mounting issue before. It is because linked slave creatures for whatever reason never go into evade mode when leaving combat. Will forward it to killerwife again, but we may need to do the same workaround as I did for the Stormwind Cavalrymen here: https://github.com/cmangos/tbc-db/commit/991eef0d5735f8ea41d297dcec559f63dce527c9

AnonXS:
Same Behavior can be observed with the Nightelf patrol between Dolanaar and Darnassus.

PR #195, opened 2017-10-09
[ACID] Terrorguard 18661

AnonXS:
http://www.wowhead.com/npc=18661/terrorguard


Phatcat:
This one should also know dual wield..

billede

PR #209, opened 2017-10-21
[ACID] Vacillating Voidcaller 19527

AnonXS:
http://www.wowhead.com/npc=19527/vacillating-voidcaller

* not sure about arcane missle spellid
* probably should have pathing with endscript spellid http://www.wowhead.com/spell=34842/de-materialize and despawn
* not sure about phase 1 shadow bolts
* not sure about spell 34302, probably linked to 34842.


Saltgurka:
There is also this spell:

> Id: 34524
> Name: Elemental Response
> Description: This generic aura self-triggers a different spell for each school of magic that lands on the wearer: Fire: 34192 Frost: 34193 Arcane: 34194 Nature: 34195 Shadow: 34196 Holy: 34197 Physical: 34198 Use this spell ID# on the creature to perform the appropriate action.

Not sure exactly how that would work and if we need to use that spell then I think it has to be done in core.

Saltgurka:
Some research is still needed for this NPC and its spells, but for now this is a good start. Pushed in https://github.com/cmangos/tbc-db/commit/c71b5ba8b5bb08612494121a2cd1e755c37312ea

PR #232, opened 2017-10-21
[ACID] Bloodmaul Mauler 19993

AnonXS:
http://www.wowhead.com/npc=19993/bloodmaul-mauler

20-23-29-21-21-28-25-28-27-20-24-30 -> 8-14,20-30 for rage, remove Buzz


Saltgurka:
These texts are done randomly when he casts Bloodmaul Rage:
image
Bloodmaul Rage should also be repeatable and can be cast at any health %.
 (you can base timers on the timestamps of the texts in image above)

Bloodmaul Buzz should not be cast in combat. It is an OOC event which should be done on waypoints. This is not yet scripted but would be done in dbscripts instead when it is.

PR #257, opened 2017-10-22
[ACID] Ridgespine Stalker 20714

AnonXS:
http://www.wowhead.com/npc=20714/ridgespine-stalker


Saltgurka:
Should probably have aura 3616 instead (it procs spell 744).
Also should have aura 7939 and 22766.

All of them should be safe to add to creature_template_addon instead of acid

AnonXS:
What Aura is 7939 ?

Saltgurka:
Dummy sneak aura that was made serverside in tbc. http://db.vanillagaming.org/?spell=7939

Don't know why they have it, but they do

PR #267, opened 2017-10-22
[ACID] Markaru 20775

AnonXS:
http://www.wowhead.com/npc=20775/markaru

unbenannt

- [x] Missing Serverside for Summoning Entry 21335 -> 36626

36664,37956 maybe


Saltgurka:
He should also summon these: http://www.wowhead.com/npc=21335/tentacle-of-markaru#abilities

Unsure exactly how that mechanic should work

AnonXS:
prob again some serverside spell we need to search for.

AnonXS:
Added Serverside and Script

PR #305, opened 2017-10-25
[ACID] Painmistress Gabrissa 21309

AnonXS:
http://www.wowhead.com/npc=21309/painmistress-gabrissa

https://web.archive.org/web/20130530055305/http://www.wowhead.com:80/npc=21309#abilities

Should use OOC Channel 32783 on Entry 21310, but this seems to not be visually correct with `spell_script_target`, the "beam" is missing.


Saltgurka:
Issue with the OOC channel is probably that the trigger npc is inside of the gameobject. Probably solved the same way as here: https://github.com/cmangos/mangos-tbc/commit/96a2c9612492ae75eef9d2f47fa91e2bee8a05a0

PR #331, opened 2017-10-28
[ACID] Hawkbane 21724

AnonXS:
http://www.wowhead.com/npc=21724/hawkbane


evil-at-wow:
I don't remember ever seeing this one, so I checked and apparently Hawkbane was removed in patch 2.1, which I guess also explains why he wasn't part of the 'Bloody Rare' achievement during WoTLK. 
Do we have spawns for this one in TBC-DB?

Now, regardless of the fact that this NPC shouldn't be spawned in TBC-DB, we can still add the abilities of course, for people that have some custom tweaks and want a spawn anyway or something.

AnonXS:
we dont have a spawn, but i figured that it doesnt harm to have it.

PR #353, opened 2017-10-29
[ACID] Cavern Crawler 22044

AnonXS:
http://www.wowhead.com/npc=22044/cavern-crawler


Saltgurka:
Altough we have no way to confirm it, proc aura 3616 should probably be used here

AnonXS:
done

PR #388, opened 2017-10-31
[ACID] Brood of Anzu 23132

AnonXS:
http://www.wowhead.com/npc=23132/brood-of-anzu

https://wow.gamepedia.com/index.php?title=Brood_of_Anzu&direction=next&oldid=2919820


Saltgurka:
What is your source for the Screech spell? I can only find this one being cast by them in sniffs:

```
Id: 40240
Name: Spite of the Eagle
ToolTip: Deals 300 damage every 1 second.
```

AnonXS:
Just have the gamepedia article to back it up, but as i left it out on l4g also i think i came to the conclusion that, from memory/checking other databases, they didnt have this ability.

AnonXS:
Closing until evidence is found that this is legit.

PR #398, opened 2017-10-31
[ACID] Insidion 23281

AnonXS:
http://www.wowhead.com/npc=23281/insidion

https://web.archive.org/web/20121016234747/http://www.wowhead.com:80/npc=23281#abilities

* Might use 9573 and 40032 only in special cases

- [ ] Skettis/Ogrila/Netherwing Ledge


AnonXS:
https://github.com/cmangos/tbc-db/commit/31c89827d8e48597921d57b4fe5dcfeea5865021

PR #399, opened 2017-10-31
[ACID] Obsidia 23282

AnonXS:
http://www.wowhead.com/npc=23282/obsidia

https://web.archive.org/web/20120226143520/http://www.wowhead.com:80/npc=23282#abilities

* Might use 9573 and 40032 only in special cases

- [ ] Skettis/Ogrila/Netherwing Ledge


AnonXS:
https://github.com/cmangos/tbc-db/commit/31c89827d8e48597921d57b4fe5dcfeea5865021

PR #417, opened 2017-11-01
[ACID] Dark Iron Land Mine 8035

AnonXS:
https://github.com/cmangos/issues/issues/1544

---

http://www.wowhead.com/npc=8035/dark-iron-land-mine

```
.go c id 6212
```

- [ ] AI Issues due to GuardianAI / Summons / TotemAI (20343)

* Current Petsystem (Spell Summoned Pets/Totems) doesnt allow all ACID Scripts to be applied
* Issues with executing IDs:

---

803501: Doesnt execute ACTION_T_COMBAT_MOVEMENT, but does execute ACTION_T_AUTO_ATTACK

---

803502

to

803504: Does not execute ACTION_T_TEXT_NEW, but does execute ACTION_T_SET_PHASE

---

803505: Does execute

Spawning the npc without summoner makes everything work out as intended. .npc add 8035

---

Might need Despawn instead of Die for Event 803505

### DIFF
--- a/ACID/acid_tbc.sql
+++ b/ACID/acid_tbc.sql
@@ -23684,8 +23684,8 @@ INSERT INTO `creature_ai_scripts` (`id`,`creature_id`,`event_type`,`event_invers
 ('621102','6211','9','0','100','1','0','5','1900','12100','0','0','11','8374','1','0','0','0','0','0','0','0','0','0','Caverndeep Reaver - Cast Arcing Smash'),
 ('621103','6211','8','0','100','0','9798','-1','0','0','0','0','1','-31','0','0','0','0','0','0','0','0','0','0','Caverndeep Reaver - Emote on Radiation Spellhit'),
 ('621104','6211','2','0','100','0','15','0','0','0','0','0','25','0','0','0','1','-47','0','0','0','0','0','0','Caverndeep Reaver - Flee at 15% HP'),
--- Dark Iron Agent (3.3.5a Official Data)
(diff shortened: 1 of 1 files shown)


Phatcat:
> Might need Despawn instead of Die for Event 803505

Here's a video showcasing them despawning/blowing up a bit;

https://youtu.be/slewhK-p8Ag

Notice the land mine that the boar attacks at around 0:58, it despawns immediately.
Also the land mine at 1:42 doesn't actually get to explode but despawns due to.. either it's master evading or itself leaving combat, it's hard to tell - the one which explodes actually leaves behind a corpse for a short while and can be seen biefly at 1:02 (notice the boar taking fire damage and the land mine's red name turning gray)

AnonXS:
Hmmm so it preexpload when its not even armed yet, "emote" or what should be eone?, that that it would even work fully atm...?

Phatcat:
If it is killed, then it despawns without a corpse, if it explodes, it will have a corpse for a brief moment (don't know for how long), that's all I was trying to say, I was just trying to answer your open question on how it behaves when it explodes, which seems to be leaving a corpse behind, unlike if it is killed before exploding.

Does it need to be scripted to die? does the explosion spell not take care of that?

AnonXS:
you're right i guess Effect 1 of spell 4043 might take care of it's (de)spawn behavior, so its just bandaide partly !yes!. i think the ability activation on contact might be another thing but thats probably more linked to how "pets/guardians" work in general somehow idk.

PR #418, opened 2017-11-01
Add Aura for Rekktilac 8277

AnonXS:
http://www.wowhead.com/npc=8277/rekktilac


Updates/xxx_c8277_aura.sql
```diff
@@ -0,0 +1,4 @@
+-- Remove aura 13787 from Bloodmaul Soothsayer c.22384
+-- Now handled in ACID since creature addon is not suitable for auras without infinite duration
+UPDATE creature_template_addon SET auras=NULL WHERE entry=22384;
```
evil-at-wow: That creature entry and comment do not match the commit message and Rekk'tilac's entry. Could this be a copy/paste error? Did you forget to save the file?

AnonXS: ty

PR #426, opened 2017-11-01
[ACID] Correct Spellid for c.22342

AnonXS:
http://www.wowhead.com/npc=22342/deathshadow-spellbinder

https://web.archive.org/web/20110625084603/http://www.wowhead.com:80/npc=22342


berserkingyadis:
I tried this PR and it fixes the Spellbinders morphing into gnomes. They now cast Arcane Missles.

PR #430, opened 2017-11-02
Fixed loot condition for Murkblood Escape Plans from Sludge Covered Object

berserkingyadis:
This PR fixes the item_loot_template of http://www.wowhead.com/item=32724/sludge-covered-object so that it can drop http://www.wowhead.com/item=32726/murkblood-escape-plans

The condition is: 
`(Has Minimum Rank Friendly With Faction ID: 1015 AND (Quest ID 11081 NOT Taken AND NOT Rewarded AND Quest ID 11075 Rewarded))`

source: https://wow.gamepedia.com/Murkblood_Escape_Plans


Saltgurka:
Thanks! Pushed here: https://github.com/cmangos/tbc-db/commit/094812fcce4e13ecd2e5de1c74cfca42f0bf3ebf :)

PR #436, opened 2017-11-03
[ACID] Umbrafen Slavebinder 18042 

AnonXS:
http://www.wowhead.com/npc=18042/umbrafen-slavebinder


AnonXS:
should target only the druid, will be added to core.

PR #437, opened 2017-11-04
Added missing quest end RP for quest 703 and quest 9719 should be available for both factions.

Golfbyxan:
Check and see if the dbscript_string entry is free.


Saltgurka:
Thanks! Pushed in https://github.com/cmangos/tbc-db/commit/69eb36c1a62d7af4b602a83c7463aee6a875f601 and https://github.com/cmangos/tbc-db/commit/89377b0d2afc99175d89a12f009e79c2eb3706bc

PR #506, opened 2017-11-10
[ACID] Altar of Shatar Vindicator 21986

AnonXS:
http://www.wowhead.com/npc=21986/altar-of-shatar-vindicator


Saltgurka:
What is your source for the net spell?

AnonXS:
http://www.wowhead.com/npc=21986/altar-of-shatar-vindicator#comments
```
Curse You Blizzard! These guys have nets, which allows them to stop you from escaping them after they decide to kill you for no reason.
```

Seems all at least tbc guards have that spell.

PR #515, opened 2017-11-11
[ACID] Daggertail Lizard 22255

AnonXS:
http://www.wowhead.com/npc=22255/daggertail-lizard

* Cant aggro when invisible


Saltgurka:
Source for Warp and Warp Charge? Wowhead only shows Phase Burst and Swipe 31279

AnonXS:
https://web.archive.org/web/20110803234543/http://www.wowhead.com/npc=22255#abilities

Swipe maybe substituted something there

PR #561, opened 2017-11-28
[ACID] Add Remove Permanent Feign Death (sleep animation) for SHH NPCs

AnonXS:
Shattered Halls Sleeping Npcs (should be dynamic pooled = more AI additions), Aura confirmed in Sniff.

https://youtu.be/z2pZgpA1cjU?t=9m10s


AnonXS:
wrong

PR #564, opened 2017-12-10
[ACID] Snarlflare 14272

AnonXS:
http://www.wowhead.com/npc=14272/snarlflare

https://web.archive.org/web/20110703120442/http://www.wowhead.com:80/npc=14272#abilities

Abilities could be added in Wrath or later or at the end of TBC when the new wave of rare npcs was added to the world.


AnonXS:
unbenannt

AnonXS:
@AnonXS  Remove this.

PR #578, opened 2017-12-18
[ACID] Ragepaw 14342

AnonXS:
http://www.wowhead.com/npc=14342/ragepaw

https://web.archive.org/web/20110625133103/http://www.wowhead.com:80/npc=14342#abilities


AnonXS:
By dionfyrelyt (2,006 –  1·22)  on 2007/07/21 (Patch 2.1.3) |
Standard melee mob, no special abilities at all and hits for very little. Can also confirm 25 Timbermaw rep

unbenannt

PR #579, opened 2017-12-18
[ACID] Alshirr Banebreath 14340

AnonXS:
http://www.wowhead.com/npc=14340/alshirr-banebreath

https://web.archive.org/web/20111023203455/http://www.wowhead.com:80/npc=14340#abilities


AnonXS:
unbenannt

PR #603, opened 2018-03-23
fix Raid Loot.

keyshuwen:
--It may need to be improved.


AnonXS:
Thank you for checking in with this.

This will be looked at and evalutated(% changes) when the export of the missing data (001_TBC_Raid_Boss_Loot_Revamp.sql) is happening, which is somewhen in spring, late spring when more other data will get exported too.

AnonXS:
The valid contents of this pullrequest will be upated in an upcoming export file. 
Thanks for providing this temp fix though.

PR #611, opened 2018-07-10
Update Enslaved Netherwing Drake inhabit type.

BillytheOrc:


BillytheOrc:
https://www.youtube.com/watch?v=jOtutz3JMRI

For documentation.

BillytheOrc:
If you do accept this PR, you can cherrypick the hacky part out. I manually modified the walkspeed of the drake rather than doing it via script.

AnonXS:
Hey there, 

thanks for trying to report back some useful data. 
This is more an issue reminder than a fix for the actual issues at hand.

This PR trys to resolve issues which have their roots in the core, by inserting false correct data.

`InhabitType` should never be modified in script as its a general notifier, which shows a creatures movement pattern/capability.

`SpeedWalk`, `SpeedRun` currently only support one generalized speed value for flying units, so it's a conversion issue between the sniffed speed values and how they are processed in the core.

---

Also in my opinion `SpeedWalk`, `SpeedRun` should only be accepted in this form:

``` `SpeedWalk` = (sniffed value / 2.5) ```
``` `SpeedRun` = (sniffed value / 7) ```

---

There are alot of Issues where there is no 100% accuracy needed to fix the actual issue at hand and which are not considered hack-fixes on the spot. I'd recommend concentrating on these issues if you do not have sniff data to lookup on certain data.

Else you can come in contact with us on [Discord](https://discord.gg/6Qjfea) so we can talk & work things out / get you more involved if you like.

Thank you!

PR #615, opened 2018-07-18
Fix Captain Saeed quest credit on gossip select.

BillytheOrc:


AnonXS:
More info on what quest this is used in to fasten up the validation would be very welcome.

Commented text needs a space between -- and the text. (See [build error](https://ci.appveyor.com/project/cmangos/tbc-db-smax9/build/1.6.2280)

Also we add 2 new lines at the end of file due to ugly no newline message.

PR #627, opened 2018-10-29
Added missing emote.

bdebaere:
There's a discrepancy between this value in classic-db and in tbc-db. Needs extra confirmation before approving!


Saltgurka:
There are several more of these creature missing the emote so rather fix all of them at once instead of just one. It could even be converted to a creature_template_addon unless there are some that should not do the emote

Grz3s:
@Saltgurka 
https://github.com/cmangos/wotlk-db/commit/9393b9f2c262821f1278cb576055c6ed3c84b792

PR #636, opened 2019-01-27
remove unused custom_data folder

Phatcat:
Not in use since the introduction of InstallFullDB.sh over the long since removed Windows_Install_Script.BAT (which must have happened years ago)

The Custom_Data folder also serves little to no purpose even if re-introduced into InstallFullDB.sh since the dev folder serves exactly the same function.


AnonXS:
The custom data folder is for projects which want to use custom content. It exists to prevent custom data updates from corrupting the normal cmangos-tbc-db updates by seperating them.

Phatcat:
So what exactly makes it different to the dev folder? Also when are the updates in custom_data applied when using InstallFullDB.sh?

Why are you the only db that has this folder? - It is not present on classic-db or wotlk-db

AnonXS:
custom data should never be included in normal cmangos-db setup, should always be applied afterwards, if someone wants to have it applied the person is free to edit their install scripts, but the gist is that custom content is not officially supported by cmangos and should always be kept seprate from the normal db setup.

Phatcat:
> should always be applied afterwards

From the readme of the dev folder;

> All updates in this directory will be applied AFTER all regular updates (DB, core, SD2, ACID...)

So again, how is it functionality wise different from the dev folder? Why supply them with a folder that they manually have to add the updates for or edit the installscript, when they could by that point easily add their own folder as well? 

From the readme from the custom data folder;

> When you use option #1 in the Windows_Install_Script.BAT it will also look for any custom data here and include it when it updates the Server.

So just leave behind something which is clearly old and outdated, unneeded and unused because 'it allows for custom content', while it does so only with configurations that are needless with the use of the dev folder, and the removal of the custom_data folder does not hamper or prevent custom content in any way...

Levitanious:
I'd say "change the readme", "include off-by-default option in script", rather than just remove the folder.

PR #645, opened 2019-03-09
Questline fix for exodar call of water

res1n:
Questline was missing exp values. Pulled values from http://web.archive.org/web/20070514144939/http://www.thottbot.com/q9500

Let me know if the format needs to be different, I was only able to get XP actually rewarded when setting the QuestLevel variable correctly, -1 would never reward the experience.

Closed previous PR due to my screwup on rebase.


AnonXS:
Resolved by https://github.com/cmangos/tbc-db/commit/e2bcb82c3140edd301231a9c2edcabaa8c4bead9

PR #646, opened 2019-04-18
Can Not Interrupt Hurricane From Druid And NPC

Gelori:
Those two spells were totally missing InterruptFlags,
NPC could continue to cast even mooving, added standard
15 flag
Issue:
   https://github.com/cmangos/issues/issues/1865


AnonXS:
spell_template changes are done in core repo.

https://github.com/cmangos/mangos-tbc/blob/master/sql/base/dbc/cmangos_fixes/Spell.sql

Gelori:
ok ty! will change it than

PR #650, opened 2019-05-26
Change chess event creatures health to 100%

BenDol:
Resolves https://github.com/cmangos/issues/issues/1912


AnonXS:
`SELECT * FROM `creature` WHERE `curhealth` != 0; ` hmmm something is wrong and missing in OS, nice catch!

BenDol:
@AnonXS I'm curious is everything suppose to have a curhealth 0 in the db? There are no cases where a creatures curhealth is not supposed to be equal to its max hp upon start up?

AnonXS:
Well it was decided to drop this always setting maxhealth as curhealth value madness, instead the core should handle to set npcs to 100% hp on spawn by default.
Scripted Cases where npcs should not have maxhp are handled by their scripts, not the curhealth column which can be nuked then at some point.

evil-at-wow:
@BenDol There are indeed cases of NPCs in the world that shouldn't have their maximum health. One such example is Tavara in TBC, that Draenei priests have to heal in the quest "Help Tavara" as part of their priest/healing introduction. Similar quests exist for other races.

For some background info: as AnonXS mentions, the curhealth column in the creature table is somewhat problematic. Apart from a certain redundancy because of maxhealth, there's also the issue that a creature spawn is often not a fixed level but in a level range (see MinLevel and MaxLevel in creature_template), which influences its maxhealth and thus typically also it curhealth values. So typically spawns don't even have a fixed curhealth value...

PR #651, opened 2019-06-01
Remove duplicate creatures from AV

BenDol:
Resolves https://github.com/cmangos/issues/issues/1885


AnonXS:
Please no Copy Paste PRs from nonblizzlike sources like https://gitlab.com/_Velvet_/AtlantisDB/commits/master .

Every change made here has to be evaluted, which is not nice work if its a random paste of code from npcs in sunwell to undercity to av.

creature_ai_x updates should be merged into https://github.com/cmangos/tbc-db/blob/master/ACID/acid_tbc.sql

BenDol:
Okay no problem, I just figured having something was more Blizzlike than nothing haha

PR #653, opened 2019-06-09
_0898_Quest_9629.sql

Elmo337:
-- Quest 9629 "Catch and Release" Murloc Tagger worked but dont Count Tagged Murlocs


AnonXS:
**Missing Spelltargeting Export**

Unbenannt

https://www.wowhead.com/quest=9629/catch-and-release#comments

ToDo Notes:

- Should add UNIT_FLAG_IMMUNE_TO_PLAYER 256 on 30877 spellhit, 5sec Delayed Despawn.

Problem:
- Baseentry is taken as viable target for the item so its possible to spam the item on the same already transformed npc https://www.wowhead.com/npc=17654/tagged-blacksilt-scout when the despawn time is increased.

- spell_dummy_npc executes too slowly (1-2secs delay) making it possible to spam the item on the same target more often.

PR #655, opened 2019-06-17
Fix WSG flag positions based on sniffs

BenDol:
Thanks @AnonXS


berserkingyadis:
New positions visualised in https://github.com/cmangos/mangos-tbc/pull/301#issuecomment-502550312

PR #791, opened 2020-06-22
Fix for the quest Morbent Fel

stealthgun:
This Pull Request fixes the quest Morbent Fel.

This will fix https://github.com/cmangos/issues/issues/2215

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ InstallFullDB.config
 
 # hum big boss file that we hopfully have to ignore.
 Killerwife_Windows_Install_Script.bat
+/.vs
--- a/Updates/0249_morbent_fel.sql
+++ b/Updates/0249_morbent_fel.sql
@@ -1,2 +1,2 @@
---Fix for the quest Morbent fel
+-- Fix for the quest Morbent fel
 update quest_template SET ReqCreatureOrGOId1=1200 where entry = 55;
--- a/Updates/0249_morbent_fel.sql
+++ b/Updates/0249_morbent_fel.sql
@@ -1,0 +1,2 @@
+--Fix for the quest Morbent fel
+update quest_template SET ReqCreatureOrGOId1=1200 where entry = 55;


Grz3s:
this fix is correct only for ClassicDB (its set like this there anyway)..
unfortunately is not correct for TBC and WOTLK
... 
Quest was reconstructed in TBC ... and Morbent Fel after using item on him suposed to be transformed into Weakened Morbent Fel = npc:24782
Classic version was backported to TDB and WDB prob by mistake ... and thats the reason why it doesnt work anymore.

Heres some info from wowhead:
`By 43854 (558 – 6·7) on 2008/01/30 (Patch 2.3.3) 
Nothing special anymore, after you have cleared the building from normal mobs use the offhand and he turns into Weakened Morbent Fel (http://www.wowhead.com/?npc=24782) and is like any other lvl32 caster. You can also use the stairs/walls to line of sight the bolts he casts`

`By 80660 (536 – 7) on 2008/10/16 (Patch 3.0.2)
A heads up, If you don't weaken him before killing him you won't get credit for your quest. Also, make sure you have good LoS with him or the weakening won't work but still activate the CD.`

quest_template for q=55 is correct there and should not be changed...

stealthgun:
So this quest is working on classic?

cala:
Last time I checked, it was working on Classic core. But as I discovered today, while checking waypoints, many scripts that were previously working and are now at least partly broken, I can’t tell for sure.

stealthgun:
> Last time I checked, it was working on Classic core. But as I discovered today, while checking waypoints, many scripts that were previously working and are now at least partly broken, I can’t tell for sure.

Is this an SD2 or DB script?

cala:
DBScript if memory serves.

stealthgun:
In the core i found:

`
        case SPELL_SACRED_CLEANSING:
        {
            if (uiEffIndex == EFFECT_INDEX_1)
            {
                if (pCreatureTarget->GetEntry() != NPC_MORBENT)
                    return true;

                pCreatureTarget->UpdateEntry(NPC_WEAKENED_MORBENT);
                return true;
            }
            return true;
        }`

in the file spell_script.cpp (https://github.com/cmangos/mangos-tbc/blob/master/src/game/AI/ScriptDevAI/scripts/world/spell_scripts.cpp)

The Id is correct but it just seems to not work.

cala:
Hell… This code is also in Classic and should not be there.  -_-"

stealthgun:
Closing because not the correct fix

Grz3s:
@cala could you remove that ancient code from all 3 cores pls...
Classic im sure works correct 
fix for wotlk (can be backported into tbc) added: https://github.com/cmangos/wotlk-db/commit/bddcd11a1e1a24b347d9364fcc6f3d2e79d6ca9b

So script its not req anymore.
+
`UPDATE creature_template SET ScriptName = '' WHERE entry = 1200;`

PR #793, opened 2020-06-22
Fix for the quest The Skettis Offensive

stealthgun:
Has been tested and fixes the server crash


stealthgun:
Closing because not the correct fix

PR #794, opened 2020-06-23
Fix for fel cannons moving

stealthgun:
This PR will fix the fel cannons moving and following you around (wich they should not do).

all speed is set to 0 so they are stationary


Grz3s:
invalid - sory

PR #817, opened 2021-03-02
Fix Faerlina's missing door

insunaa:
In conjunction with https://github.com/cmangos/mangos-tbc/pull/432

No door existed previously.
The GUID was chosen by the game server and is thus a "generated" guid, I don't know how to pick a proper GUID, please advise or change.

PR #818, opened 2021-03-02
Add Faerlina outer ring door

insunaa:
https://github.com/cmangos/tbc-db/pull/817#issuecomment-788860303
Added fix to the correct location and used proper GUID

Only makes sense together with https://github.com/cmangos/mangos-tbc/pull/432 otherwise the door will always be closed, as the wrong GUID was used in the instance_naxxramas SD2 script.


insunaa:
No idea why that `1,` was there.
I added the GO ingame and extracted the SQL command from the `2021-03-02_logSQL.sql` file and it is present there, as well.

Should work properly now, but I guess it needs to be double-checked

cala:
Merged. Thank you! 👍

PR #829, opened 2021-03-08
Fix warning message that command doesn't exist

insunaa:
At some point the syntax was changed from `.combatstop` to `.combat stop`, but this field in the Database wasn't updated to reflect this change and now it is throwing warnings when the server is started.


al3xc1985:
@cyberium this can be added 2 to all cores cause it makes a lot of errors... it is not something needed to be tested especially by a db dev

cyberium:
👍 will do a bit later

cyberium:
Done core side thanks for your contribution

PR #898, opened 2021-03-17
Fix Thunderclap speed reduction

insunaa:
Thunderclap used by Sethekk Guardians (in Sethekk Hall) used to reduce movement speed by ~99% in Heroic and ~80% in non-heroic, where it should instead be 40%

Fixes https://github.com/cmangos/issues/issues/2419


AnonXS:
@killerwife

AnonXS:
resolved by core change, that cuts a "double amount effect application" for this type of spell.

PR #938, opened 2021-06-06
Hide Event Triggers for Non-GMs

shagu:
Fixes: https://github.com/cmangos/issues/issues/2469
Thanks @killerwife 

Removing player visibility for the following objects:
- 128972 Zul'Farrak Graves
- 176750 Desolace Kodo Graveyard Skulls
- 171941 Blackrock Keg Trap
- 180391 Heart of Hakkar Spell Emitter
- 176592 Shellfish Trap
- 177493 Fire of Elune
- 177529 Altar of Elune
- 178124 Resonite Crystal
- 178248 Naga Brazier
- 181214 Necropolis critter spawner
- 184718 Cauldron Summoner
- 184722 Cauldron Bug Summoner
- 103575 Containment Coffer TRAP

Identified via:
`SELECT * FROM gameobject_template WHERE type = 6 AND displayId = 327 AND data8 != 1`

PR #939, opened 2021-06-18
Hide GO Traps for non-GMs

shagu:
Removes the visibility of those objects:
image

Same as already previously suggested here: https://github.com/cmangos/tbc-db/pull/938 
I also added the "Frostwolf Landmine" and "Stormpike Landmine" objects. 

The [solution](https://github.com/cmangos/tbc-db/commit/a0d2500e65e74972197881a2c6ad7eff281b775a) that was applied instead does not work because it uses the `data9` field. Using `data8` as in my suggestions hides those objects to non-GM characters.

@AnonXS

### DIFF
--- a/Updates/0481_hide_gotraps.sql
+++ b/Updates/0481_hide_gotraps.sql
@@ -1,0 +1,18 @@
+-- Hide Gameobject Traps to player characters.
+-- Removing player visibility for the following objects:
+--   128972 Zul'Farrak Graves
+--   176750 Desolace Kodo Graveyard Skulls
+--   171941 Blackrock Keg Trap
+--   180391 Heart of Hakkar Spell Emitter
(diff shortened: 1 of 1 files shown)


shagu:
Are my pull requests not welcomed, did I made something wrong or is there any other reason why you refuse to simply click the merge button on my PRs?

insunaa:
The PRs are merged through git amends, not through the github interface. The reason is so that the changes can be replicated across the three different cores/databases.

That's just the workflow, it has nothing to do with the quality of your commits.

shagu:
> The PRs are merged through git amends, not through the github interface

So the workflow is to force-push on every change? :eyes: 
Thanks for explaining. I'm glad I didn't made anything wrong.

Issue #679, opened 2019-09-23
Quest The shattered salute 2460 - startscript


killerwife:
Questgiver explains saluting

miraco:
You can btw close this issue @killerwife

PR #975, opened 2021-12-02
Add quest_start_script for Quest The Shattered Salute(2460)

miraco:
Dont have sniffs, base on video: https://www.youtube.com/watch?v=A5xh6_LLakM
https://github.com/cmangos/tbc-db/issues/679

### DIFF
--- a/Updates/01_The_Shattered_Salute.sql
+++ b/Updates/01_The_Shattered_Salute.sql
@@ -2,7 +2,8 @@ INSERT INTO `broadcast_text` (`Id`, `Text`, `ChatTypeID`, `VerifiedBuild`) VALUE
 ('77866', '$n, pay very close attention to what I do and follow my instructions precisely. To not heed my words is to sign your own death warrant, as Taskmaster Fizzule is VERY well versed in the ways of the warrior and the assassin.', '0', '18019'),
 ('77867', 'After you get his attention - and you can be certain that if he sees you near the tower, you\'ll have his undue attention - face him and perfom the Shattered /Salute. Let me demonstrate....', '0', '18019'),
 ('77868', 'Shenthul prepares the Shattered Salute! *PAY ATTENTION!*', '2', '18019'),
-('77869', 'Did you see what I did? If you are unsure of the technique used in presenting the /salute, practice on me. If you correctly execute the Shattered /Salute, I will give you the proper response.', '0', '18019');
+('77869', 'Did you see what I did? If you are unsure of the technique used in presenting the /salute, practice on me. If you correctly execute the Shattered /Salute, I will give you the proper response.', '0', '18019'),
+('77870', 'Well done!', '0', '18019');
 
 INSERT INTO `dbscripts_on_quest_start` (`id`, `delay`, `priority`, `command`, `datalong`, `dataint`, `comments`) VALUES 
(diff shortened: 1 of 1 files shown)


killerwife:
That video is from a pserver. Please look at this one https://www.youtube.com/watch?v=wcKUNwe0HFg

miraco:
Okay, im gonna update the script. 
In the first sentence he calls the name of the player, i only know $N to get player name, but i get an error cause the talk dont get send to player.
is there a way to set the player as target/buddy in dbscripts?

killerwife:
Yes, its possible to do that.

miraco:
I updated the script now. 
The part after the player /salutes has to get changed in core. 
Timers may need an update (if someone has sniffs)

> Shenthul prepares the Shattered Salute! *PAY ATTENTION!*
This part could also be with %s instead of hardcoded name.

killerwife:
Actually it doesnt need a core change. https://github.com/cmangos/issues/wiki/Creature_ai_scripts here check EVENT_T_RECEIVE_EMOTE or am I missing smth.

killerwife:
Oh I see, his script is in core. That script can easily be removed and done in EAI. Through condition it can be checked for having quest and the talking there as well. Do you want to try it? Just need to disable SD2 by setting shenthuls scriptname in creature_template to ''

miraco:
Its currently scripted in core, in file orgrimmar.cpp
But yes we can set it into an AI script aswell. Gonna try it then.

miraco:
Just one question to it, currently its scripted that the Quest fails after 60 seconds, is this intendet? And is this also doable via creature_ai/condition?

killerwife:
I will let you know in 10 minutes if its supposed to fail.

killerwife:
I can indeed confirm that the 60 seconds fail is not blizzlike. So only the emote reaction is needed in EAI.

miraco:
Okay then this should work. 
Can we keep it in this file, or do we have the eventAI part into acid.sql?

killerwife:
You dont need to worry about that. However what needs to be noted is that the broadcast_text entries for all of this already exist and the existing ones need to be used and emote/chattype updated in them. This is a fairly recent addition. I will fix this and link to you the final product.

miraco:
Okay, already found them:
3278
3279
3280
3281
3287

killerwife:
I did a few  alterations and pushed it. Please look at what I did for future reference. Appreciate the contribution. Should you want to help on more, between me and @ratkosrb we have data for all of them which do not have a video linked in it. Just write on the issue you wanna look at and I can provide data. Some are already linked to vmangos and only need a port.

miraco:
Yeah i will try to do some more.
Just one last question to this:
The EventAI for this is repeatable and i just tested ingame, you can spam /salute macro and he will try to response all the time. Could this be a way to crash the Server/make it laag ingame? Cause we have more AI scripts that response to an emote for example:
`Chief Thunder-Skins respond to emote Dance with Dance id = 2514903`

All these creatures could get emote spammed and make world server laag.

Hope you understand what i mean xD

killerwife:
The relay script can only be running once, however the fix here would be to set the condition value2 to 1. You can try that. What it does is that it checks if quest is taken but only evaluates as true if objectives arent complete.

Issue #961, opened 2021-09-29
Quest 3379 - Shadoweaver - missing afterscript


killerwife:
The big infernal says a text either as afterscript or as he despawns.

killerwife:
Also check if raze gives staff of command for controlling him.

ratkosrb:
https://github.com/vmangos/core/commit/3e35c126448d6738e5894f9e5f14edf12c5c79a2

miraco:
https://github.com/cmangos/tbc-db/commit/b1329af4d9d8890dd6a1b6f33c0c4d8604961100

PR #978, opened 2021-12-03
Add completion script for quest Shadoweaver.

miraco:
https://github.com/cmangos/tbc-db/issues/961

Based on https://github.com/vmangos/core/commit/3e35c126448d6738e5894f9e5f14edf12c5c79a2

This looks kinda wierd tbh. He gets despawned, and respawns after 5secs? Also currently she plays some emotes when you finished the quest and open gossip. "OfferRewardEmote = 1" for example.

Never worked with vmangos, is delay in vmangos like in cmangos? or is delay 3 = 3 seconds, so we have to use 3000 in cmangos?


ratkosrb:
Yes, delay uses seconds in vmangos, while its milliseconds in cmangos. And yes, that is what happens when you turn in the quest. The npc says something, despawns, then respawns after 5 seconds.

Issue #964, opened 2021-10-08
Quest 2204 - Restoring the Necklace - missing afterscript


killerwife:

killerwife:
https://github.com/vmangos/core/commit/fb9a2a23dec5b061fde238937039a2e3bae132be

miraco:
https://github.com/cmangos/tbc-db/commit/618ba8814b28c7806143a5e1591b42346195431f

PR #980, opened 2021-12-03
Add completion script for quest Restoring the Necklace.

miraco:
https://github.com/cmangos/tbc-db/issues/964

Based on https://github.com/vmangos/core/commit/fb9a2a23dec5b061fde238937039a2e3bae132be

But for me its kinda wierd there are no orientations for the 2 waypoints. 
The result is the npc stands just next to the books/bed:
http://puu.sh/ItnDH/aff4ff66ac.jpg
http://puu.sh/ItnDr/6805f3e116.jpg

I guess vmangos made it with sniffs or is this just missing there?


Grz3s:
I would also add `21 SCRIPT_COMMAND_SET_ACTIVEOBJECT` with correct flag on master... to be 100% sure that script will go from A to Z ... instead of stucking in the middle (koz of its duration time) .. in case if player for any reason disappear from there... That would cut sctipt leaving npc  without npc flags or even in diff place koz command 3 is also used.
npc should be active for ehole script and just before end  make him Unactive (this is also v. important)
look on this one for example:
https://github.com/cmangos/wotlk-db/commit/ddcc62f42f11494de78decc12a8e6fc56b9aacf9#diff-72f679a0623f4acbbcac9a84946de969a426e21dcb58ec007434a79507245bacR5

miraco:
Yeah I will add it. Only reason player would disappear from this would be a disconnect, because you only get the item after the event from an follow up quest. 

For future should we add this to every script that is longer then 10 seconds for example?

Grz3s:
`For future should we add this to every script that is longer then 10 seconds for example?`
im adding it to everything that may be canceled for X reasons... player disconnect is one of them,,,
also if 3rd party is used (another X number of npcs) .. i would 100% sugest to make them all active ..

killerwife:
The orientation is not needed if you set a waypoint orientation to 100, that will make the npc keep the orientation as it was moving towards its destination.

Issue #704, opened 2019-10-12
Quest Forbidden Knowledge 737 - afterscript


killerwife:

killerwife:
https://github.com/vmangos/core/blob/development/sql/migrations/20211201050831_world.sql

miraco:
https://github.com/cmangos/tbc-db/commit/76048df9de287eaca0fab7e2cdca52a2755220b2

PR #985, opened 2021-12-04
Add completion script for quest Forbidden Knowledge.

miraco:
https://github.com/cmangos/tbc-db/issues/704
Based on https://github.com/vmangos/core/blob/development/sql/migrations/20211201050831_world.sql

maybe someone else know the solution for this:
http://puu.sh/ItCTK/4f797f0a5c.jpg
On chat it uses npc name, in chat bubble not.


miraco:
The reason was simple, its an emote not a chat bubble. :)

Issue #705, opened 2019-10-17
Quest Solution to Doom 709 - afterscript


killerwife:
Misses some RP.

killerwife:
https://github.com/vmangos/core/blob/development/sql/migrations/20211201050010_world.sql

miraco:
https://github.com/cmangos/tbc-db/commit/beb1495c7dcb0cc953c1e3018fb0971f8ca107f5

PR #986, opened 2021-12-04
Add completion script for quest Solution to Doom

miraco:
https://github.com/cmangos/tbc-db/issues/705
Based on https://github.com/vmangos/core/blob/development/sql/migrations/20211201050010_world.sql


AnonXS:
grafik

👍

Issue #892, opened 2021-03-15
Quest 7461 - The Madness Within - afterscript


killerwife:
These spawn and do stuff https://www.wowhead.com/npc=14364/shendralar-spirit#comments

ratkosrb:
https://github.com/vmangos/core/commit/d062e86d56025b9505e6b3e7d870e92860a23e99

miraco:
This looks a bit tricky to add in cmangos.
Vmangos has the option to let the spawned creature have an individual script.
If i understand this script right, 8 npcs with id 14364 get spawned, and all of them should handle an differnt script.

Do we have a way to do this in cmangos? 
SCRIPT_COMMAND_TERMINATE_SCRIPT uses entry, so every npc will use it.

Then i changed my local code to:
![ss+(2021-12-06+at+03 10 42)](https://user-images.githubusercontent.com/6261245/144860949-3d6de683-38b5-4b5d-80a4-a3b4854300ed.png)
Imo we can add relayScript as dataint 4 so we can specify different relay scripts for spawned creatures with same entry id.

@killerwife  whats your opinion to this?

killerwife:
@miraco how is the npc spawned?

miraco:
They get spawned via db_quest_end_script. They dont move, just change facing to the player, and do some emote.

killerwife:
Ok, I know what we need for this, I will code a feature real quick, rework some existing stuff and ping it here.

miraco:
> Ok, I know what we need for this, I will code a feature real quick, rework some existing stuff and ping it here.

Okay, just in case:
https://github.com/vmangos/core/commit/0f8ce721da882766f33e499ae7997f7a10802d1e

For this script 
NPCs get spawned via db_quest_end_script. And after some time they will move and cast a spell.

killerwife:
https://github.com/cmangos/mangos-tbc/commit/806bcbfaca32ee19fca6b4d30f08c5ab39936db8
https://github.com/cmangos/tbc-db/commit/526fb9c97f2023c65e8b124ccb6e42bab51636e7
https://github.com/cmangos/tbc-db/commit/ebb8294e554c173a7089b715806fe25ea029672a

Note: numbering scheme for spawn data template - vanilla uses 100-9999 range, 10000+ is for tbc, this is so we dont have overlap.

miraco:
Thanks for your work.

[...]

PR #993, opened 2021-12-06
Add completion script for quest The Madness Within.

miraco:
https://github.com/cmangos/tbc-db/issues/892
Based on https://github.com/vmangos/core/commit/d062e86d56025b9505e6b3e7d870e92860a23e99

Have to recheck sniffs/videos i guess.


ratkosrb:
Here is a video of the quest from sniff replay. You can see they all face the player a second after they spawn.
https://streamable.com/tfzdfp

Issue #967, opened 2021-11-01
Quest 2521 - To Serve Kum'isha - missing afterscript


killerwife:
In vmangos

killerwife:
https://github.com/vmangos/core/commit/0f8ce721da882766f33e499ae7997f7a10802d1e

miraco:
https://github.com/cmangos/tbc-db/issues/892#issuecomment-986813369
Same problem same entryid spawned multiple times with different relay script.

Nytanath:
Fixed by https://github.com/cmangos/tbc-db/pull/994 I believe?

miraco:
Yes

PR #994, opened 2021-12-07
Add completion script for quest To Serve Kum'isha.

miraco:
https://github.com/cmangos/tbc-db/issues/967

Based on https://github.com/vmangos/core/commit/0f8ce721da882766f33e499ae7997f7a10802d1e

This may will need some more updates.

- [x] recheck ids for relay scripts. 9999+ normaly is set for TBC, currently using questid*10 for movement scripts (25210, 2521 and so on)
- [x] NPCs are hovering during movement. (mmaps maybe wrong for this spot?)
- [x] May clear the sql file a bit 
- [x] Spawn Position seems to be wrong.  Screenshot from retail TBC

145051435-2da03c4c-8b9b-42db-bcc6-840dfa85abe8

For the last point, i dont wanna change spawn coords with guessed coords. Maybe someone have sniffs/can sniff his exact coords.

### DIFF
--- a/Updates/00021_quest_to_server_kumisha.sql
+++ b/Updates/00021_quest_to_server_kumisha.sql
@@ -1,13 +1,10 @@
-DELETE FROM dbscripts_on_quest_end WHERE id = '2521';
-INSERT INTO `dbscripts_on_quest_end` (`id`, `delay`, `command`, `datalong`, `datalong2`, `datalong3`, `data_flags`, `dataint`, `comments`) VALUES 
-('2521 ', '0', '45', '2521', '0', '0', '0', '0', 'To Serve Kum\'isha: - Start RelayScript');
-UPDATE `quest_template` SET `CompleteScript`=2521 WHERE `entry`=2521;
-
 -- Main Script
-DELETE FROM dbscripts_on_relay WHERE id = '2521';
-INSERT INTO `dbscripts_on_relay` (`id`, `delay`, `priority`, `command`, `datalong`, `datalong2`, `datalong3`, `buddy_entry`, `search_radius`, `data_flags`, `dataint`, `dataint2`, `dataint3`, `dataint4`, `x`, `y`, `z`, `o`, `comments`) VALUES 
-('2521', '0', '0', '21', '1', '0', '0', '0', '0', '0','0', '0', '0', '0', '0', '0', '0', '0', 'To Serve Kum\'isha: Kum\'isha the Collector - Set ActiveObject'),
-('2521', '104000', '0', '29', '3', '2', '0', '0', '0', '0','0', '0', '0', '0', '0', '0', '0', '0', 'To Serve Kum\'isha: Kum\'isha the Collector - Remove Npc Flags'),
+DELETE FROM dbscripts_on_quest_end WHERE id = '2521';
+INSERT INTO `dbscripts_on_quest_end` (`id`, `delay`, `priority`, `command`, `datalong`, `datalong2`, `datalong3`, `buddy_entry`, `search_radius`, `data_flags`, `dataint`, `dataint2`, `dataint3`, `dataint4`, `x`, `y`, `z`, `o`, `comments`) VALUES 
+('2521', '0', '2', '32', '1', '0', '0', '0', '0', '0','0', '0', '0', '0', '0', '0', '0', '0', 'To Serve Kum\'isha: Kum\'isha the Collector - Check if we are in range'),
+('2521', '0', '1', '21', '1', '0', '0', '0', '0', '0','0', '0', '0', '0', '0', '0', '0', '0', 'To Serve Kum\'isha: Kum\'isha the Collector - Set ActiveObject'),
+('2521', '0', '2', '32', '1', '0', '0', '0', '0', '0','0', '0', '0', '0', '0', '0', '0', '0', 'To Serve Kum\'isha: Kum\'isha the Collector - Pause Waypoints'),
+('2521', '104000', '0', '29', '3', '0', '0', '0', '0', '0','0', '0', '0', '0', '0', '0', '0', '0', 'To Serve Kum\'isha: Kum\'isha the Collector - Remove Npc Flags'),
 ('2521', '111000', '0', '0', '0', '0', '0', '0', '0', '0','3441', '0', '0', '0', '0', '0', '0', '0', 'To Serve Kum\'isha: Kum\'isha the Collector - Say Text'),
 ('2521', '112000', '0', '10', '7364', '15797', '0', '0', '0','0', '0', '0', '0', '0', '-11366.8', '-2979.6', '-0.198364', '0.628319', 'To Serve Kum\'isha: Summon Creature Flawless Draenethyst Sphere'),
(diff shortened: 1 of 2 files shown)


al3xc1985:
for second point try to use normalize Z = 1 in mangosd.conf it should work
I encountered the same problem in the even start zone on elves and that was the solution to the problem, to activate that feature. it uses more cpu but it helps

ratkosrb:
The whole script is auto generated from sniff. All positions in it are correct.

miraco:
Not talking about the script points, talking about the quest giver itself, and you can see in screenshot from tbc retail he has other spawn position then he has in cmangos

Grz3s:
> @ratkosrb 
> The whole script is auto generated from sniff. All positions in it are correct.

Sor for offtopic.. 
Just wonder.. how that auto generator would work (would it handle them?) with little bit harder scripts like this one.. where one reaction depends from another one.. and depends from what outcome is -- ending script may have few diff variants..
heres example:
https://www.youtube.com/watch?v=dXFxM4-1tPo&list=PLQWa1A2Qh_1XH_dyeGPZE1D9hwPw801e1&index=21

ratkosrb:
The tool simply generates script commands from events in the sniff you select. It does not know what is triggered by what. First I rewatch the sniff in the replay core to actually see what is happening. Then i use the tool to generate all the sql for the scripts, and i assign them to the relevant places myself. It's primary use is with scripted sequences that require no interaction from the player. Just like the quest this pull request is for. To manually write all that sql code would be wasting time, instead the tool can do it for you with the click of a button, and this way you don't risk making mistakes.

Here is what the tool looks like.
Untitled

You select the events you want included in the script from the list, and then click the "Make Script" button and it spits out the sql inserts. It does not generate EventAI events, only dbscripts.

ratkosrb:
For the example in the video you posted, I'd use it to make a script for everything that happens after the last npc dies, and I would attach it to the Death event of the npc in EventAI myself. First I would add just the last npc to the object filter, and search for Health Updates, to get the time of death. Then I'd copy that time, and set it as the start time, and make it list events for Admiral Barean, so i can have it turn that scripted sequence at the end into a dbscript which can be started from EventAI.

Grz3s:
@ratkosrb thanks for your explanation...
Youre right... tool is very nice for simple scripts like the one from this topic.. however.. it misses checks (easy done with dbscripts):
- prevent for multispawn - in case is smth is summoned after q completed - and few ppl does it in same time. 
- check if creature (buddy system) exists. - in case if third party npc/creature is used in event.. and is not there for any reason
- longer events should be active (loaded for whole time) - in case if someone will not stay there or disconnect... event will cut in middle?
I would sugest to think about it ... with upcoming scripts .. to prevent tons of errors in console just because of that.
BTW:
Vid i did show.. is my completed work, so does not need anything else .. it was just example..

ratkosrb:
> - prevent for multispawn - in case is smth is summoned after q completed - and few ppl does it in same time.
> - check if creature (buddy system) exists. - in case if third party npc/creature is used in event.. and is not there for any reason

This is for the human using the tool to consider, and figure out whats the most appropriate manner to do it. The tool is just for automated generation of script commands from events in sniff. In the case of this script, this is prevented by the fact the first thing I do on quest completion is start a scripted map event, and I have the "Abort on Failure" flag assigned. So if 2 people turn it in at the same time, the script will abort for the second person, since a scripted map event with this ID is already running. It's impossible for it to run more than once simultaneously and cause duplicated spawns.

> - longer events should be active (loaded for whole time) - in case if someone will not stay there or disconnect... event will cut in middle?

I have not seen such a problem on vmangos. Scripts run on Map::Update, so they will continue executing as long as the map exists. A missing target, like a player logging out during the script, will not interrupt script execution. Just the commands that target him will throw an error, which is what we want to happen, but it will continue executing the rest of the script, and any npcs involved should still have their state correct in the end. If you want a script to abort on error, there is an "Abort on Failure" flag that can be assigned to individual commands.

killerwife:
Note on the cut in middle, this is due to grid unload. I suspect you are running vmangos with it turned off by default. Servers generally do not unload grids for this reason.

al3xc1985:
@killerwife what should be the best setting on cmangos when we talk about this grid? The optimal settings?

killerwife:
For development and cmangos internal purposes always grid unload turned on. Its mainly a memory saving method. Servers with 16+GB ram dont need it rly.

miraco:
So i finally had sometime to watch over my open PRs again.

https://youtu.be/-ejg963Izfo 

For this i got his waypoints he uses every ~450. Only thing i now need to find out:
how this quest event gets handled when he is moving since he spawns the objects and castes spells only in the crater.

I would guess he only starts the quest_end_script when he is either in range of the gameobject (id 137167) or if he is on waypoint/spawn point (in this case waypoint 22)

Nytanath:
I just tested this. When you turn in the quest while he's far from the spawnpoint, once the waypoint pause due to talking with him expires, he runs to the middle of the crater and starts the event. After the event ends, he seems to get stuck. I've waited more than 20 minutes for him to start his waypoints again, but nothing.

Nytanath:
https://mega.nz/file/miYHTQJK#UKVBDTrP_D4LUpviS1ub8PkE_BOJnDq_GHXlEarhqT0

I expected it to be viewable from Mega, but alas, it doesn't seem to work. Sorry about that.

Highlights:
04:40: turned the quest in.
06:40: npc loses questgiver flag, starts running to center of crater.
06:41: talked with the NPC to see if he'd stop again. He did.
08:41: NPC resumes running towards center of crater. Rest of the quest event follows.
From 10:12 onwards: I tried to see how long it'd take for him to start his out of combat waypoints. That turned out badly, as the 30 following minutes show him being stuck there. I kept the client open for another 20 minutes while the video was processing - nothing.

What's funny is that you could technically grief other players with the way it works - wait until the NPC is very far from crater, turn quest in, wait until he loses questgiver flag and keep talking to him, triggering the 2-minute WP pause. Boom, no more quest giver for a while.

miraco:
Thanks, this helps me realy much. I will look into this when i have finished the other quests im currently working on.
but after watching the video i only think wtf ist blizz doing with this.. ill think about doing something "custom" here so the npc starts his waypoints again after quest event happend

Nytanath:
> wtf ist blizz doing with this

I often think exactly that too :D

If you need more tests let me know, on the PTR I can reliably solo the dragon that drops the item for this quest at a 100% chance.

miraco:
> > wtf ist blizz doing with this
> 
> I often think exactly that too :D
> 
> If you need more tests let me know, on the PTR I can reliably solo the dragon that drops the item for this quest at a 100% chance.

It seems like he never do his waypoints after someone end this quest.  I watched it 3 times now on ptr and the npc just isnt moving anymore. On shadowlands he doesnt have his waypoints anymore and just use rnd movement at the crater. Maybe this got patched in 2.4 already.

Ill let this open for now and think about making it either custom, using normal waypoints and on quest end run to crater but starting waypoints after the event again. 
Or just remove waypoints again. 
Realy strange npc behavior

killerwife:
There are plenty of resetting issues on classic. Take the Red vs Blue or Mulgore brawl fights. Dont think about it too much and make it actually work on our end repeatedly.

PR #996, opened 2021-12-08
Start working on quest the binding start script

miraco:
I started working on creating the StartScript for quest "The Binding"
Based on https://github.com/vmangos/core/commit/ba8e80f7703758e8eb3871391fef85c251dedcca#.
Somethings still missing and i currently dont have an idea how to solve it.

- [ ] The Binding: Strahad Farsan - Cast Spell Summon Effect - sometimes the creature is just staying in the room after her waypoints and dont cast the Spell.
- [ ] Need a solution for scripting the Failure/Success event.

For 2nd part, vmangos use 'SCRIPT_COMMAND_START_MAP_EVENT' with condition 'CONDITION_ESCORT' to activate Sucess (17950) or Failure (17951) script.
Sucess happens when player killed the npcs.
Failure either happens when player dies or the npcs in the room (?). 

My thoughts was to add event ai to the Felhunter that starts an RelayScript on Death/EVENT_T_KILL.
But we atleast would need to check if the player that got killed from the npc had the quest or it was just an rnd player. Else it would fail when npc insta kills an level 1 player in range for example.

Maybe you have an idea for this @killerwife 

video of it: https://www.youtube.com/watch?v=EOt6uVMi5F4&t=366s

### DIFF
--- a/Updates/0022_quest_the_binding.sql
+++ b/Updates/0022_quest_the_binding.sql
@@ -1,0 +1,66 @@
+-- Start script for quest The Binding.
+DELETE FROM dbscripts_on_quest_start WHERE id = '1795';
+INSERT INTO `dbscripts_on_quest_start` (`id`, `delay`, `command`, `datalong`, `datalong2`, `datalong3`, `data_flags`, `dataint`, `x`, `y`, `z`, `o`, `comments`) VALUES 
+('1795 ', '0', '21', '1', '0', '0', '0', '0', '0', '0', '0', '0', 'The Binding: Strahad Farsan - Set ActiveObject'),
+('1795 ', '1000', '29', '51', '2', '0', '0', '0', '0', '0', '0', '0', 'The Binding: Strahad Farsan - Toggle Npc Flags'),
+('1795 ', '1000', '1', '1', '0', '0', '0', '0', '0', '0', '0', '0', 'The Binding: Strahad Farsan - Emote OneShotTalk'),
+('1795 ', '1000', '0', '0', '0', '0', '0', '2370', '0', '0', '0', '0', 'The Binding: Strahad Farsan - Say Text'),
+('1795 ', '4000', '3', '0', '0', '0', '0', '0', '-780.438', '-3718.21', '42.1855', '0', 'The Binding: Strahad Farsan - Move'),
+('1795 ', '6000', '3', '0', '0', '0', '0', '0', '-769.382', '-3721.35', '42.4199', '0', 'The Binding: Strahad Farsan - Move'),
+('1795 ', '14000', '15', '8677', '0', '0', '4', '0', '0', '0', '0', '0', 'The Binding: Strahad Farsan - Cast Spell Summon Effect');
+
+UPDATE `quest_template` SET `StartScript`=1795 WHERE `entry`=1795;
+
+-- Script when Strahad Farsan casts spell Summon Effect
+DELETE FROM dbscripts_on_event WHERE id = '1428';
+INSERT INTO `dbscripts_on_event` (`id`, `delay`, `command`, `datalong`, `datalong2`, `datalong3`, `buddy_entry`, `search_radius`, `data_flags`, `dataint`, `x`, `y`, `z`, `o`, `comments`) VALUES 
+('1428 ', '0', '9', '29205', '120', '0', '0', '0', '0', '0','0', '0', '0', '0', 'The Binding: Summon GameObject Strahad\'s Summoning Circle'),
+('1428 ', '2000', '1', '2', '0', '0', '6253', '30', '0', '0', '0', '0', '0', '0', 'The Binding: Acolyte Fenrick - Emote OneShotBow'),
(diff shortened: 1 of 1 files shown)


ratkosrb:
I am not sure cmangos has a way of doing quests like this in the database. I use "Scripted Map Events", which is a feature I came up a with a few years ago so that I can script escort quests or anything else that requires managing the state of multiple objects. The event runs on the map, separately from any units that might be part of it, and can have success and failure conditions and scripts attached. It also has a time limit, after which it will automatically fail if not completed. You can store targets in the event, and also assign success/failure conditions and scripts to each one individually. I do that here when the felhunter is summoned. I add him to the event and attach a (Not Is Alive) success condition to it, so that the event finishes when it dies. This feature has made it possible to script many quests in the database, which would have required c++ scripts otherwise.

In the case of this quest, the event has the following properties.
- Fail after 5 minutes expire, in case player goes afk after accepting quest.
- Fail if the player goes more than 50 yards away from the quest giver.
- Fail if the player dies.
- Fail if the quest giver dies.
- Succeed once the felhunter demon summoned at the end dies.
- On failure, forcibly respawn all the npcs involved in the script, and despawn the summoned felhunter demon.
- On success make the quest giver walk back to his spawn position and restore his npc flags.

Grz3s:
make sense... as i said im looking at it from cmangos perspective...
BTW: 
`On success make the quest giver walk back to his spawn position and restore his npc flags.` - you not addin this to escoretd npcs :) im sure noone would like them to walk back 😄

miraco:
Yeah thats what i thought.
Maybe someone will add it 'Scripted Map Events' or something simular to cmangos one day. 
@killerwife ;)

@Grz3s Good that this is not an escort quest.

killerwife:
I will take a look at this in a bit of a greater detail soon™. Havent been able to do much today.

ratkosrb:
@Grz3s Those points I listed are specifically what it's used for on this quest. The npc should walk back once you kill the felhunter.

al3xc1985:
@ratkosrb it would be nice if @killerwife implement this feature in cmangos to update vmangos on cmangos style, to have a simplier porting between projects, what do you think?

miraco:
Just leaving this here https://github.com/vmangos/core/commit/265420f1e2b25159214ee5da056b3decd4b0ed54#diff-f64feb50ea458af6421de727a36ce96bf323f8e2d1ea02bdfcc48a5cedf67bb0

ratkosrb:
I've implemented the [feature](https://github.com/ratkosrb/NostalriusTBC/commit/0a7b25e48f95ff18101e6f2fec15fe6d4531f403) on my cmangos fork, but it's up to cmangos developers to decide if they are okay with this functionality being added to dbscripts.

ratkosrb:
I forgot to initialize the events timer :smile: 
Pull the latest commit and try again. Also added the escort condition.

miraco:
Will do when im Home again. 
I saw you also respawn the 3 acolytes ok failure. On my local the Felhunter never attacked the quest giver or one of his friends, is there another reason why you also respawned the acolytes and not just the Quest giver?

ratkosrb:
Just to restore their npc flags and make sure they will be in place, cause they are part of the script. Creatures are reset to their default state and position on respawn.

miraco:
Okay, works good for me. Thanks!

al3xc1985:
Vwry well @miraco that is super nice

miraco:
I finally found some time to work at this again. 
This works now without the extra code from vmangos.

Succes strict gets triggert when felhunter is dead.

Failure script gets triggert when farsan is still standing near his acolyte (in the house) after 5 minutes.

only difference to vmangos is.
On fail i let the npc just move outside to his spawnpoint and add his npc flag instead of respawning it. 

not sure whats blizzlike, dont think anyone has every tried to fail this quest.

miraco:
I finaly managed to do this Quest on PTR to see Failure scripts.

Failure Script gets triggert when:
- 3 minutes after object got spawned and nothing happend
-  when felhunter is out of combat longer then 6 seconds (or 6 seconds after he killed a player)

When the "event" failed quest npc starts moving back to his spawn point and gets questgiver flag back same for acolyts.

A video from TBC ptr: https://youtu.be/4A5T8e_BAkc

miraco:
i reworked the sql file for this for now. i still dont wanna merg this cause there is to much happening.

One thing i find kinda strange, on quest end, questgiver npc (strahad farsan entry 6251) removes questgiver flag. (2)
This means he has npc flag 49 instead of 51 (like in sniffs)
`ServerToClient: SMSG_UPDATE_OBJECT (0x27C4) Length: 1618 ConnIdx: 1 Time: 02/14/2022 23:15:31.846 Number: 302235
[0] Object Guid: Full: 0x204CB00020061AC0001FA3000005659D Creature/0 R4908/S8099 Map: 1 Entry: 6251 Low: 353693
[0] UNIT_NPC_FLAGS: 49`

This should be 
UNIT_NPC_FLAG_TRAINER_CLASS
UNIT_NPC_FLAG_TRAINER
UNIT_NPC_FLAG_GOSSIP

When using npc info after accepting the quest it still shows the quests but he has npc flag 49
Wow_pNRBoTL8E6

PR #1004, opened 2021-12-30
Create  0056_missing_enchantment_item.sql

danjohnso:
Was looking through the enchantment vendor and noticed he was missing Assault to Bracer, Agility to gloves and stealth to cloak enchants.  The items do not exist in the item_template table, and so far my efforts to add them have failed in that the items are purchasable, but not usable.  I wanted to contribute what I had so far, maybe someone who understands the database better can comment as to what I am missing to make the items usable.  Also, the displayids are impossible to track down, they don't match the icon ids at all and I am at a loss as to how to get proper icons on these item_templates.


AnonXS:
PTR stuff https://tbc.wowhead.com/npc=26302/enchantments-vendor

add space after -- to make built not break

danjohnso:
> PTR stuff https://tbc.wowhead.com/npc=26302/enchantments-vendor
> 
> add space after -- to make built not break

PTR/Tournament realm

MantisLord:
I think it's a good idea if someone wants to add missing item templates using actual blizz data from some source, as was done here https://github.com/vmangos/core/commit/eb6ee47247835f1878775d2a0cb19ed1120a156e

But adding entirely custom item ids/data is probably outside the scope of this project, sorry.

PR #1011, opened 2022-01-05
Add spawn_group data for hellfire ramparts

miraco:
I will just leave this here for now.
This is still in the making and i have to learn more about the new spawn_group system. 

Just to have an example how it could look like in the future.

I added all groups to the new system, but had to outcommit some cause i am not sure how to handle them correct. 

Biggest benefit for me is the better movement from groups. 

ToDo
- [x] Figureing out how the groups with 1 Destroyer and 1 npc left from him (darkcaster/archer) and 1 npc right from him (darkcaster/archer) currently handled by pool system.
- [x] Single NPCs that dont belong to groups (To get  them also use dynamic guid ingame i guess)
- [x] Fix DB Errors
- [x] Maybe add some more comments.
- [ ] Beastmaster missing random movement after.
- [x] distance (FormationSpread)

Will do some more work for this after work/tommorow.

### DIFF
--- a/Updates/Instances/543_hellfire_ramparts.sql
+++ b/Updates/Instances/543_hellfire_ramparts.sql
@@ -15,23 +15,23 @@ SET @SGGUID := 54300; -- spawn_groups
 -- CREATURES
 -- =========
 
-INSERT INTO `creature_movement` (`id`, `point`, `position_x`, `position_y`, `position_z`, `orientation`, `waittime`, `script_id`) VALUES
+INSERT INTO `creature_movement` (`id`, `point`, `PositionX`, `PositionY`, `PositionZ`, `orientation`, `waittime`, `ScriptId`) VALUES
 -- (@CGUID+2, 1, -1270, 1642.56, 91.6092, 3.55975, 0, 0),
 -- (@CGUID+2, 2, -1232.97, 1662.94, 92.4019, 0.464499, 0, 0),
 -- (@CGUID+4, 1, -1337.58, 1676.92, 91.7461, 2.14132, 0, 0),
 -- (@CGUID+4, 2, -1358.19, 1701.99, 84.6861, 2.40209, 0, 0),
 -- (@CGUID+4, 3, -1337.58, 1676.92, 91.7461, 2.14132, 0, 0),
 -- (@CGUID+4, 4, -1324.93, 1661.38, 93.0835, 5.34654, 0, 0),
-(@CGUID+17, 1, -1276.92, 1640.51, 91.6726, 2.69188, 22000, 1173),
-(@CGUID+17, 2, -1276.52, 1638.6, 91.6777, 4.62397, 0, 0),
-(@CGUID+17, 3, -1278.69, 1636.47, 91.6955, 3.85742, 0, 0),
-(@CGUID+17, 4, -1282.16, 1637.86, 91.706, 2.761, 0, 0),
-(@CGUID+17, 5, -1282.55, 1639.29, 91.7032, 1.84051, 0, 0),
-(@CGUID+17, 6, -1282.15, 1640.54, 91.6982, 1.23026, 26000, 1173),
-(@CGUID+17, 7, -1283.46, 1638.84, 91.7092, 4.3852, 0, 0),
-(@CGUID+17, 8, -1281.31, 1635.72, 91.7092, 5.35595, 0, 0),
-(@CGUID+17, 9, -1277.05, 1636.46, 91.6883, 0.177039, 0, 0),
-(@CGUID+17, 10, -1276.07, 1639.22, 91.6735, 1.16271, 0, 0),
+-- (@CGUID+17, 1, -1276.92, 1640.51, 91.6726, 2.69188, 22000, 1173),
+-- (@CGUID+17, 2, -1276.52, 1638.6, 91.6777, 4.62397, 0, 0),
+-- (@CGUID+17, 3, -1278.69, 1636.47, 91.6955, 3.85742, 0, 0),
+-- (@CGUID+17, 4, -1282.16, 1637.86, 91.706, 2.761, 0, 0),
+-- (@CGUID+17, 5, -1282.55, 1639.29, 91.7032, 1.84051, 0, 0),
+-- (@CGUID+17, 6, -1282.15, 1640.54, 91.6982, 1.23026, 26000, 1173),
+-- (@CGUID+17, 7, -1283.46, 1638.84, 91.7092, 4.3852, 0, 0),
+-- (@CGUID+17, 8, -1281.31, 1635.72, 91.7092, 5.35595, 0, 0),
+-- (@CGUID+17, 9, -1277.05, 1636.46, 91.6883, 0.177039, 0, 0),
+-- (@CGUID+17, 10, -1276.07, 1639.22, 91.6735, 1.16271, 0, 0),
 -- (@CGUID+19, 1, -1315.75, 1601.66, 91.7501, 4.33494, 0, 0),
 -- (@CGUID+19, 2, -1305.9, 1593.47, 91.7652, 5.82091, 0, 0),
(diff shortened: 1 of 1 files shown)


Updates/Instances/543_hellfire_ramparts.sql
```diff
@@ -9,18 +9,19 @@ EndDBScriptData */
 SET @CGUID := 5430000; -- creatures
 SET @OGUID := 5430000; -- gameobjects
 SET @PGUID := 48200; -- pools
+SET @SGGUID := 54300; -- spawn_groups
```
miraco, line 12: Using SG for "spawn_group"

Updates/Instances/543_hellfire_ramparts.sql
```diff
@@ -163,81 +218,225 @@ REPLACE INTO `creature_template_addon` (`entry`, `mount`, `bytes1`, `b2_0_sheath
 (17455, 0, 0, 1, 0, 0, '18950'), -- Bonechewer Beastmaster
 (17517, 0, 0, 1, 0, 0, NULL); -- Hellfire Sentry
 
+INSERT INTO `spawn_group`(`Id`, `Name`, `Type`, `MaxCount`, `WorldState`, `Flags`) VALUES 
+(@SGGUID+1, 'Helfire Ramparts - Group 1', 0, 0, 0, 3),
+(@SGGUID+2, 'Helfire Ramparts - Group 2', 0, 0, 0, 3),
+-- (@SGGUID+3, 'Helfire Ramparts - Group 3', 0, 3, 0, 3), -- Bonechewer Destroyer and Archer/Darkcaster
```
miraco, line 224: will need more information to let this group work like its working now.

Updates/Instances/543_hellfire_ramparts.sql
```diff
@@ -163,81 +218,225 @@ REPLACE INTO `creature_template_addon` (`entry`, `mount`, `bytes1`, `b2_0_sheath
...
+(@SGGUID+27, @CGUID+3, 1), -- Bonechewer Hungerer
+(@SGGUID+28, @CGUID+33, 0), -- Bleeding Hollow Darkcaster
+(@SGGUID+28, @CGUID+34, 1), -- Bleeding Hollow Darkcaster
+(@SGGUID+29, @CGUID+4, 0), -- Bonechewer Hungerer
+(@SGGUID+29, @CGUID+5, 1), -- Bonechewer Hungerer
+(@SGGUID+30, @CGUID+99, 0), -- Hellfire Sentry
+(@SGGUID+30, @CGUID+100, 1), -- Hellfire Sentry
+
+INSERT INTO `spawn_group_formation`(`SpawnGroupID`, `FormationType`, `FormationSpread`, `FormationOptions`, `MovementID`, `MovementType`, `Comment`) VALUES 
+(@SGGUID+4, 2, '2.2', 0, @CGUID+20, 2, 'Bonechewer Ravener'),
```
miraco, line 358: got some errors when using InstallFullDB.sh again, but no errors when just importing it directly to the database.

AnonXS: yea because spawn_group_formation is not part of the instance clear file yet

miraco: Nah it was much easier, forgot a ; in line 355 :<

miraco:
Just as an example how im taking the 'FormationSpread' currently:

For the groups with 1 leader and 2 followers (for example here ravener and 2 warhounds) im going on the spawn position from warhound (.go c 5430065) and do .dist with the moving npc as target same with the other warhound.

Taking the raw distance from both sides and then get the half. So for this example:
Wow_Gp6Lu5ydLM
Wow_JdVxebsebo

(3,19 + 3,62) / 2 = 3,405. So i took 3.4 for this group.

Maybe we can find a better solution for this in the future.

miraco:
For the groups with 1 Destroyer staying in front, and archer/darkcaster behind him:

Wow_SZU2kqcPE6

`SET @CGUID := 5430000; -- creatures
SET @SGGUID := 54300; -- spawn_groups

INSERT INTO `spawn_group`(`Id`, `Name`, `Type`, `MaxCount`, `WorldState`, `Flags`) VALUES 
(@SGGUID+3, 'Helfire Ramparts - Group 3', 0, 3, 0, 3); -- Bonechewer Destroyer and Archer/Darkcaster

INSERT INTO `spawn_group_spawn`(`Id`, `Guid`, `SlotId`) VALUES
(@SGGUID+3, @CGUID+51, 0), -- Bonechewer Destroyer (gets always spawned)
(@SGGUID+3, @CGUID+27, 1), -- Bleeding Hollow Darkcaster or Archer
(@SGGUID+3, @CGUID+40, 2); -- Bleeding Hollow Darkcaster or Archer

INSERT INTO `spawn_group_entry` (`Id`, `Entry`, `MinCount`, `MaxCount`, `Chance`) VALUES 
(@SGGUID+3, 17269, 0, 2, 0),
(@SGGUID+3, 17270, 0, 2, 0);

SET @CGUID := 5430000;
DELETE FROM creature WHERE guid IN (@CGUID+27, @CGUID+102, @CGUID+40, @CGUID+101);
INSERT INTO `creature` (`guid`, `id`, `map`, `spawnMask`, `position_x`, `position_y`, `position_z`, `orientation`, `spawntimesecsmin`, `spawntimesecsmax`, `spawndist`, `currentwaypoint`, `DeathState`, `MovementType`) VALUES
(@CGUID+27, 0, 543, 3, -1273.19, 1654.71, 69.2056, 1.0472, 7200, 7200, 0, 0, 0, 0), 
(@CGUID+51, 17271, 543, 3, -1269.61, 1653.95, 69.0822, 0.785398, 7200, 7200, 0, 0, 0, 0), -- Bonechewer Destroyer
(@CGUID+40, 0, 543, 3, -1268.79, 1650.25, 69.0301, 0.488692, 7200, 7200, 0, 0, 0, 0); `

Thats what i currently tryed. But this result in either having 2 darkcaster and 1 archer or in having 1 darkcaster and 2 archer.

Adding 
`INSERT INTO `spawn_group_entry` (`Id`, `Entry`, `MinCount`, `MaxCount`, `Chance`) VALUES 
(@SGGUID+3, 17271, 1, 1, 100);`

Made 1 bonechewer spawn, but it can also result in having him spawn in the back left or back right. 

My first guess would be that spawn_entry overrides the id (entry) in creature table. Imo it should only override it when it uses id 0 (for random spawn).
Or i miss understood something here.

For Single NPCs that dont belong to groups (To get them also use dynamic guid) we may just should always create a group. 
@killerwife maybe you have more infos for these 2 problems.


Database error (when using InstallFullDB.sh got solved for now

AnonXS:
Imo dont overuse it, only use it for the movement packs and completely randomized groups.

miraco:
Then I totally misunderstood something here. I thought we gonna drop creature_linking and pool system and only use spawn_groups for all creature groups.

cyberium:
Its new stuff we have some table renaming, maybe some feature to add and bug to fix, in that sense AnonXS is right.
But you are right too we need some of the case you are trying to implement to be able to identify our need and eventual bugs.
Also yes i want to enforce the removing of creature_linking system. The sooner, the better it is. But no hurry.

miraco:
> Its new stuff we have some table renaming, maybe some feature to add and bug to fix, in that sense AnonXS is right.
> But you are right too we need some of the case you are trying to implement to be able to identify our need and eventual bugs.
> Also yes i want to enforce the removing of creature_linking system. The sooner, the better it is. But no hurry.

Then I would suggest to use it for alle the groups where it’s already working like it should (in this case all moving groups and most of the static groups) 
And outcommit/remove the groups that for now don’t work as wanted.

for the things that are missing/not working  for now you want an bug report/Feature suggest?

cyberium:
Oh yes ;) Any feedback would be welcome thank you in advance

Tobschinski:
I don't like you haircut.

PR #1021, opened 2022-01-27
Install script improvement

cyberium:
Script now support
- Multiple user menu
- User settings directly modifiable in the script
- Live check of all settings
- Display status of current db
- Filling clean db from scratch with normal user (mangos user if DB already exist)
- Setting mangos user, creating and filling all data from scratch with root access
- Support of user ip limitation for mangos user
- Help to identify when db need to be updated
- Update from core folder to all DB
- Some advanced feature like delete everything if need
- Edit RealmList directly from this script no need for mysql client
- help switch and some usefull switchs


InstallFullDB.sh
```diff
@@ -270,194 +918,1196 @@ then
...
+  echo
+  true
+}
+
+# Apply locales
+function apply_locales()
+{
+  if [ "$LOCALES" == "YES" ]; then
+    echo "> Trying to apply locales data (May take some minutes) ..."
+    for UPDATEFILE in ${ADDITIONAL_PATH}locales/*.sql
```
MantisLord, line 1206: I received error like this when I try to use additional option like Locales:

```> Trying to apply locales data (May take some minutes) ...
  - Applying locales/BroadcastTextLocales.sql ... SUCCESS
  - Applying locales/OtherLocales.sql ... SUCCESS
  - Applying locales/BroadcastTextLocales.sql ... FAILED!
>>> ERROR 1062 (23000) at line 9: Duplicate entry '14746-deDE' for key 'broadcast_text_locale.PRIMARY'
Press space to continue...
```
I think it's just looping twice through the files in locales folder.

MantisLord (commented):
This is a very nice script! 👍
I tested on Windows and it works well for me.

I made one comment about the additional options trying to apply updates more than once.

I think maybe we don't need to show db versions on the main menu unless there is a mismatch. 
Also on first usage it took 5-10 seconds to create config file and such. Since there was a blank screen during this time I got a little bit confused, maybe we could print a message that says Loading or something on first launch.
These last two points are very minor and would be fine without any change though :)

cyberium:
Thank you MantisLord, Issue with broadcasttext should be fixed now.
About your second to wait before showing something i want you to retest it and see if its same now.

For everyone
I added a downloadable mysql server ready to use builded after each push. It can help some to get ready with cmangos faster.
What's your opinion about it?
You can see result [here](https://github.com/cyberium/tbc-db/actions/runs/1768959069)

berserkingyadis:
Also did a test now on my Windows 10 machine with "Git for Windows"->"Git Bash" installed. 

Generally this is a huge improvement over our current InstallScript and I like the magic going on in the background like finding the core repo and the mysql installation. It found both in my case :) Good thing is that the copy paste works also incase one needs to enter their filepaths manually.

I ESPECIALLY like that this script also now takes care of doing the initial table setup and applies the "base" sql files to each database! <3 This will be big QoL for the users and our guides need to be changed for this.

One nitpick I would have is making the disclaimer that the user is possibly deleting all their DBs and users in ALL CAPS so that they know the amount of damage they can possibly do. 

Good work, Cyberium!

MantisLord:
> Thank you MantisLord, Issue with broadcasttext should be fixed now. About your second to wait before showing something i want you to retest it and see if its same now.

Looks better now - I can confirm the locales option is fixed and I saw a message during initial load which I didn't see originally. 👍

cyberium:
Thank for your review guys.

.github/workflows/mysqlwindows.yml
```diff
@@ -0,0 +1,216 @@
...
+
+    runs-on: windows-2019
+
+    steps:
+      - name: "Set environmental variables"
+        run: |
+          echo "MYSQL_BIN_ARCH=${{env.MYSQL_BIN_NAME}}${{env.MYSQL_BIN_EXT}}" >> $env:GITHUB_ENV
+          echo "MYSQL_PREBUILD_ADR=${{env.MYSQL_BASE_PREBUILD_ADR}}${{env.MYSQL_BIN_NAME}}${{env.MYSQL_BIN_EXT}}" >> $env:GITHUB_ENV
+          $projectNames = "${{github.repository}}".Split("/")[1]
+          $expension="${projectNames}".Split("-")[0]
```
MantisLord, line 28: EXPENSION -> EXPANSION ? :)

InstallFullDB.sh
```diff
@@ -270,194 +1074,1597 @@ then
...
+  done
+  echo "  CMaNGOS fixes for DBC data were applyed successfully!"
+  echo
+  true
+}
+
+# Apply scriptdev2.sql
+function apply_full_scriptdev2_data()
+{
+  if ! execute_sql_file "$WORLD_DB_NAME" "$CORE_PATH/sql/scriptdev2/scriptdev2.sql" "> Trying to apply ScripDev2 data"; then
```
MantisLord: We need to apply Spell.sql in the scriptdev2 folder as well.

PR #1028, opened 2022-02-28
Alterac Valley tweaks

celguar:
Alterac Valley for some of issues I came across:
 - Spirit guides are missing in both Alliance & Horde caves (coordinates are from mangos-classic, mangos classic original Horde cave Spirit Guide was in 2.4+ position so I used it here)
 - Fix mine creatures links to BG events. Fix is from cmangos-classic. AV mines only have 3 states - Ally, Horde, Neutral. They don't have contested states and their event IDs depend on TeamIndex (0,1,2). Mangos-classic has these set correctly while tbc/wotlk use some contested-style event IDs (1,3,5) which don't of course.
 - Note: BG mine spawns are still not always correct, but this fix is to restore mine functionality (being captured by killing enemy unique NPC), not to fix spawns.
 - Note: I'm not sure about Spirit Healer GUIDs, I just took those close to other AV spirit guides.


AnonXS:
We have Instance File for AV in tbc-db.

celguar:
Is it in tbc-db only or in wotlk/classic too?

AnonXS:
only tbc+ as vanilla map 30 looks differently.

celguar:
Thanks! I'll make another PR then

PR #1029, opened 2022-02-28
Alterac Valley tweaks

celguar:
Alterac Valley for some of issues I came across:
 - Spirit guides are missing in both Alliance & Horde caves (coordinates are from mangos-classic, mangos classic original Horde cave Spirit Guide was in 2.4+ position so I used it here)
 - Fix mine creatures links to BG events. Fix is from cmangos-classic. AV mines only have 3 states - Ally, Horde, Neutral. They don't have contested states and their event IDs depend on TeamIndex (0,1,2). Mangos-classic has these set correctly while tbc/wotlk use some contested-style event IDs (1,3,5) which don't work of course.
 - Note: BG mine spawns are still not always correct, but this fix is to restore mine functionality (being captured by killing enemy unique NPC), not to fix spawns.


AnonXS:
I just tested this, and it seems with these values, in the alliance mine the alliance npcs are spawned by default, not the neutral ones...
I think this all got mixed up and now neither db side nor core side values are really what they should be.

celguar:
@AnonXS Did you test it with by PR for Alterac Valley fix?

celguar:
It's easy to test on clean core/db
Start AV, wait liek 5 minutes and check who owns both mines. Then try to capture any of them

celguar:
Anyway these values only work with mangos-tbc PR I made today. Without db/core fixes mines are not working as intended

AnonXS:
ah right there is core commit too no wonders^^

PR #1063, opened 2022-05-06
[RP-Event] Mordant Grimsby

miraco:
Correct movement and RP Event for Mordant Grimsby
Based on TBC PTR sniffs


AnonXS:
What about InteractionPauseTimer then?

miraco:
He doesnt stop when you speak to him while moving

https://www.youtube.com/watch?v=-QTHt6INB0I&feature=youtu.be

Do we have to set it to 0 then? (Never looked at this before
`UPDATE `creature_template` SET `InteractionPauseTimer`='0' WHERE (`Entry`='23843')`

PR #1069, opened 2022-05-25
[Quest] Call of Water

miraco:
ID 63, 96, 100

This quest looks easy, but took me some time to have an idea about redoing this.
This is my basic idea:

Let all the corrupted water elements do an despawn script (relay script) on
a) Accepting 'call of water' id 100 (gameobject, not able to do arelay script else it would need much less code)
b) on finishing 'call of water' id 1103 (backup quest if you failed it before)

Spawn objects, spawn Water Spirit with their path id.
Let water elements pause their waypoints on first waypoint.

When finishing quest Call of Water ID 96
Start relay script for Minor Manifestation of Water
let Minor Manifestation of Water threw an AI event
Water Spirit on AI_RECEIVE unpause waypoints.

To get correct movement start, add spawn waypoint with correct waittime 

Video of quest 100 
https://www.youtube.com/watch?v=8N7dL9iOj0k&feature=youtu.be

Video of Backup Quest (1103)
https://www.youtube.com/watch?v=8N7dL9iOj0k


killerwife:
And why is the part about it being a GO being a particular issue?

miraco:
It’s not an issue, but if I could start an relay script from gameobject I could use 1 relay script for both quest (100 and 1103 cause the spawn part is the same

killerwife:
I meaaaan, we can fix that easily :D

killerwife:
Delete this IF https://github.com/cmangos/mangos-tbc/blob/master/src/game/DBScripts/ScriptMgr.cpp#L2662

miraco:
If that’s ok for you, I will change it. But I guess there was a reason for not doing it, maybe because we can’t set gameobjects as active object in world so long relay scripts will stop when  no player is in Range

killerwife:
Thats also incorrect 😀

killerwife:
This line can also be changed:

https://github.com/cmangos/mangos-tbc/blob/master/src/game/DBScripts/ScriptMgr.cpp#L2104

Only thing that needs to be restricted in this one is that if it hits anything player controlled, it shouldnt go through.

miraco:
Is it also possible to use SCRIPT_COMMAND_SEND_AI_EVENT from gameobject?

killerwife:
From yes, To no. Or at least it can be adjusted because there a similar case for sending from player.

miraco:
> From yes, To no. Or at least it can be adjusted because there a similar case for sending from player.

Okay, its not necessary to have, it would just make 
`(100,0,1,45,5897,0,0,5897,18262,16,0,0,0,0,0,0,0,0,0,0,19100,'Call of Water - Corrupt Water Spirit - Start Relay Script'),`
This a bit easier, cause i could just use SCRIPT_COMMAND_SEND_AI_EVENT instead 

But its now working as it should.

All based on sniffs

PR #1070, opened 2022-05-27
Fix Crag Stalker Opener to actually use Surprise Attack

DrKittens:
This replaces the existing Enemy AI Script for Crag Stalkers so that it actually causes crag stalker to open with "Surprise Attack"

It resolves this spellcast failure spam.

``DoCastSpellIfCan by Creature (Entry: 4126 Guid: 21277) attempt to cast spell 8151 but spell failed due to unknown result 90.``

The changes to "Crag Stalker - Cast Stealth OOC" are:

action2_type = 20;action2_param1 = 0 (turns off autoattacks)

The changes to "Crag Stalker - Cast Surprise Attack" are:

action1_type = 20; action1_param1 = 1 (turns on autoattack)

action2_type = 11; action2_param1 = 8151; action2_param2 = 1 (cast Surprise Attack on target)

The toggling of autoattacks off and on seem to skew the race condition of the Crag Stalker auto-ing and breaking stealth then failing to surprise attack in favour of the Crag Stalker actually getting Surprise Attack off before the autoattack. 

This solution may also apply to other creatures who use a special opener from stealth 100% of the time that is currently failing to work.

Testing performed in sequence:

Action: Walk into Crag Stalker that is stealthed

Result: What Happens: Get hit with surprise attack then melee'd

Action: Use Feign Death effect to break combat after been hit with Surprise Attack then re-enter combat with the same creature

Result: Crag stalker re-enters stealth then after finding me again melee's me without problem

Action: Shoot a stealthed Crag Stalker with a wand to initiate combat

Result: Crag stalker melee's me but still attempts to cast Surprise Attack in the illegal state (un-stealthed) causing some console spam.

Action: AFK in range of Crag stalker

Result: Crag stalker hits me with Surprise attack, melee's me, then 30 seconds later console spam for it trying to use Surprise Attack starts up again.


DrKittens:
A better fix would be this but with an additional check on the "Crag Stalker - Surprise Attack" script to require the stealth aura from "Crag Stalker - Cast Stealth OOC" to be present before proceeding. and if its not run a third combat script that just tells it to keep melee'ing without trying to use Surprise Attack, IE combat went on for more than 30 seconds, it cant do its opener again.

I am unsure how to do this. If someone knows please let me know / ammend the script.

AnonXS:
This belongs into [ACID.sql](https://github.com/cmangos/tbc-db/blob/master/ACID/acid_tbc.sql)

DrKittens:
Fixup branch using a diverged master from upstream

Add changes into ACID.sql instead of update file.

AnonXS:
I've to say, to silence the db error for good, you have to add script in core for all spells with SPELL_ATTR_ONLY_STEALTHED to make them not try to execute when npc is not stealthed anymore. this should maybe even done generic and not by spell with the CanCastResult related to "result 90"

PR #1071, opened 2022-05-28
[Creature] Correct UnitFlags for shadowpine hexxer so they become interactive

DrKittens:
Shadowpine Hexxer (quest target creature) is non-interactable due to incorrect Unitflags been applied to it
Fixes shadowpine hexxer so it has the same UnitFlags as other shadowpine mobs it spawns with.


AnonXS:
We dont REINSERT for a _template if there is not good reason for it, why not compose it as a UPDATE?

DrKittens:
Re-composed as a cleaner update statement

Fixed up branch corruption with upstream master

PR #1072, opened 2022-05-28
[Creature] INcrease Irespeaker's buddy range

DrKittens:
Fixes console spam ``DB-SCRIPTS: Process table `dbscripts_on_creature_movement` id 2499901, command 15 has buddy 23033 not found in range 15 of searcher Creature (Entry: 24999 Guid: 5300461) (data-flags 2), skipping.``

Increments the radius to look for the Invisible Stalker (floating) buddy instead of moving either as both are visible to the player & the spell it casts is already in range, just a buddy check failing repeatedly.

Alternate fixes would involve moving either entity closer to each other but that would impact the visual spell effect used.


AnonXS:
I really dislike how this one line update looks, if so much data in one line is changed, do a reinsert for this one line so it can be reviewed more easily.

DrKittens:
Fixed branch corruption

Recomposed statement as a simpler SQL Update

AnonXS:
I can not replicate this issue, the trigger npcs are always spawned and the Irespeaker is close enough to channel the trigger.

```
18:28:05 Raw distance: 8.13
18:28:05 P -> T Combat distance: 5.63
18:28:05 P -> T Bounding distance: 7.43
18:28:05 T -> P Combat distance: 5.63
18:28:05 T -> P Bounding distance: 7.43
18:28:05 P -> T Attack distance: 44.00
18:28:05 P -> T Visible distance: 9.00
18:28:05 P -> T Visible distance (Alert): 11.22
18:28:05 T -> P Attack distance: 5.00
18:28:05 T -> P Visible distance: 45.00
18:28:05 T -> P Visible distance (Alert): 50.10
```


```
.go c 5300032
.worldstate sunsreach phase 0
```

DrKittens:
Reproduction Scenario:

1. Create a new blood elf character
2. Play the game for a few minutes / dont actually go to sunsreach

sunsreach loads in from the belf starter area and this appears to misbehave if its loaded in this fashion.

Still occurs in latest, i am using non-default settings
``GridUnload = 0`` 
``LoadAllGridsOnMaps = 1``

which may be related.

AnonXS:
Yes, but that leads me to believe that in your "default" version "range" is not the issue for these scripts failing, both the trigger npc and the Irespeaker should be spawned and active and they dont need more then ~15y range to reach each other.

DrKittens:
Agree, something else is the issue.

PR should not be merged.

PR #1073, opened 2022-05-28
[Creature] Relocate gul'dan's camera shaker buddy

DrKittens:
Fixes below console errors on AI script for gul'dan by moving the buddied invisible camera shaker creature target closer to gul'dan. Player cannot see this object unless flagged with .gm on

`DB-SCRIPTS: Process table `dbscripts_on_creature_movement` id 1700803, command 15 has buddy 21052 not found in range 30 of searcher Creature (Entry: 17008 Guid: 84600) (data-flags 7),`

### DIFF
--- a/Updates/0258_Gul'Dan_Relocate_Camera_Shaker_Buddy.sql
+++ b/Updates/0258_Gul'Dan_Relocate_Camera_Shaker_Buddy.sql
@@ -1,0 +1,1 @@
+UPDATE `creature` SET `position_x`=-3591.86, `position_y`=1845.89, `position_z`=49.5043, `orientation`=1.60497, WHERE `guid`=73950;


AnonXS:
Are you sure this location is incorrect by sniff? Also, your commit tree in this branch is corrupted.

DrKittens:
Fixed branch corruption, simplified Update statement.

The location from the sniff is probably correct, i do not have data demonstrating its incorrect the new position is likely incorrect.

The relocation was to resolve the error mentioned in the original comment.

Initially i went about this in the same manner as #1072 but when he triggered the shake script the screen continued to shake until i left the area or did /reload.

AnonXS:
As in #1072, the real issue seems to be either their random movement, position or something else needs to be investigated further.

DrKittens:
Agreed.
Closing this and #1072 as this isnt a meaningful fix and there's likely many other little things like this to address and it would be better to look into the root cause and address that.

PR #1074, opened 2022-05-28
[Reagent] Add missing alpha reagent Tiny Silver Key

DrKittens:
The 2.4.3 client is aware of this item but it is missing serverside.

This causes an error when casting the spell Khadgar's Unlocking (Rank 4) as it looks for this reagent ID 
(redtext: Missing reagent UNKNOWN)

Item has been created using data from the tbc client & wowhead, once added and given to the player the spell is castable and is functional.


AnonXS:
Can you reaffirm this data with [VMaNGOS ](https://github.com/vmangos/core)?

DrKittens:
Fixed up branch corruption.

Checking mangos.sql in VMaNGOS the spell Khadgar's Unlocking (Rank 4) also looks for Reagent ID 8148 which this creates.

VMaNGos however has a Tiny Silver Key item. Comparing the objects shows differences for display_id, buy_price, sell_price and extraflags.

I've ammended my insert to use the displayid from vmangos as it doesnt seem to make a difference to the client.

I've aligned buyprice and sellprice with vmangos after resetting my client back to vanilla to check if i had corrupted my itemcache (i had)

vmangos sets extraflags to 4, according to readthedocs for cmangos that isnt a valid type so i've left it as 0 (nothing special) same as the other tiny keys that exist in tbc-db

DrKittens:
classic-db is also missing this reagent

classic-db is additionally missing the Tiny Copper Key from tbc-db

PR #1078, opened 2022-06-02
[ACID] Murloc Netter - Add cooldown / delay to sunder armor attempts

DrKittens:
Significantly reduces below spam / server lag from Murloc Netter trying to sunder armor every single tick if the first sunder armor aura bounces tick.

``DoCastSpellIfCan by Creature (Entry: 513 Guid: 90063) attempt to cast spell 11971 but spell failed due to unknown result 8.``


AnonXS:
How does that work? https://github.com/cmangos/mangos-wotlk/blob/master/doc/EventAI.txt#L127

DrKittens:
> How does that work? https://github.com/cmangos/mangos-wotlk/blob/master/doc/EventAI.txt#L127

I am uncertain, appears i have set some totally arbitrary flags that led to it not spamming console after .reload creature_ai_scripts and respawning the mob. I'll investigate further and see what i've done. My mistake.

DrKittens:
Reverted this, restarted cmangos and db. Cant reproduce issue so assuming something else went wrong.

PR #1080, opened 2022-06-06
[Quest] Add quest script to 'Seeping corruption'

miraco:
https://github.com/cmangos/issues/issues/2860


AnonXS:
"add workaround for kill part, we dont have a 'kill self' command or set health to 0 command

* Use spell to kill npc" - are we missing a scripting capability here, normally indeed spells are used to kill off npcs.

miraco:
i didnt had a spell in sniffs when doing this quest, he just killed himself, so yeah we miss a scripting capability here.
I got the spell from [vmangos](https://github.com/vmangos/core/blob/a01720f547f0d042b7989085a1a3558cc6e92a64/sql/old_migrations/20180527162107_world.sql#L102)

insunaa:
as far as I know most scripted NPC deaths are done via spell.
Do sniffs pick up spells that are set to not appear in combat logs?

PR #1103, opened 2022-07-12
Add paymaster gossip SQL

insunaa:
This PR adds the Gossip Menu and Broadcast Text conversion for this PR:

https://github.com/cmangos/mangos-tbc/pull/536

### DIFF
--- a/Updates/0393_ptr_paymaster.sql
+++ b/Updates/0393_ptr_paymaster.sql
@@ -1,0 +1,4 @@
+UPDATE creature_template SET GossipMenuId=26751 WHERE entry=26075;
+DELETE FROM `npc_text_broadcast_text` WHERE `Id` IN (39288);
+INSERT INTO `npc_text_broadcast_text` (`Id`, `Prob0`, `BroadcastTextId0`) VALUES
+(39288, 1, 39288);


AnonXS:
I'd still pretty much like having "spawn" data for these npcs, be it part of an "arena tournament or ptr", so we can actually actively utilize these scripts.

insunaa:
There are technical limitations on the 2.4.3 Client that prevent most of the NPCs from being effectively used when filled out with accurate data (they can crash the client because the number of items the vendors sell is larger than 15 pages). KW fixed it so the client no longer crashes, but now every item past page 15 is lost to the ether.

The only real solution to make effective use of these NPCs is to make them different from the reference material (by duplicating them and splitting the vendor slots between them), which is not compatible with 1) the pedestals which summon the vendors and 2) the spawn data.

I'm implementing these like this because like this they are accurate to the source, but they can only be used with the intention of supporting the effort of whoever is testing the server.

AnonXS:
Well maybe, now after the split of vendor lists, we should "list" the other half of items in a second npc that shares the same spell in a second spell effect[x] so this all actually does something...
and we CAN use them.

insunaa:
The spawn data would still be incorrect, as the spawn data does not consider 2 NPCs being used. I can look into it, but I don't have accurate spawn data for these NPCs aside from Video evidence, which isn't particularly accurate. If someone does have the data I can make a separate PR.

AnonXS:
Nobody has spawn data of these pedestals?

al3xc1985:
> Nobody has spawn data of these pedestals? Seems strange at this point...
If anyone has a 5min afk/run around in atleast Shattrath to spare i'll add the spawn data myself.

@AnonXS this is a thing that I don't know, is it inoff to run around Shatrah, and the sniffer gets all the data? if yes I would gladly do that

miraco:
You would need to wait for tbc ptr cause those are only spawned on tbc ptr. and iirc those didnt show in any sniffs i had from shattrath on tbc ptr

al3xc1985:
@miraco can't I find them on cata in Shatrah?

miraco:
No, blizz changed their locations and despawned most of them + i think they have another size now

PR #1111, opened 2022-12-18
Some more translation corrections

Ischtaratu:
Correction of further translation errors in the deDE language


Ischtaratu:
request is outdated.

PR #1115, opened 2023-03-09
[Quest] add some cosmetic fixes for gizletone caravan quests

miraco:


AnonXS:
quite some time has already passed and inbetween expansions this file does not exist for the other repos anymore, so if its not too much do it as normal update file with update querys please.

miraco:
Done

PR #1133, opened 2023-06-01
[Karazhan] Rework Part 3 - Moroes Trash 

miraco:
Fix spawn points for all creatures
Fix movepoints for all Pats
Remove creature_linking and add it correctly into spawn_groups
Remove some RP events that get rescripted in later update


miraco:
I'll push it when im done with readding rp scripts

PR #1136, opened 2023-06-05
[Karazhan] Phantom Guest AI

miraco:
Update AI after having the newest changes to core. 
This fixes the error with spell_list not getting assigned correctly. (EventAI: Creature entry 16409 has ranged mode action but no main spell.)

On Spawn decide what Version the npc is (Phase 1-5) 
GenericTimer with 200ms delay assigns the SpellList (first) and the Melee/Range/Caster Mode. 
Not repeatable Phase 5 - Cast DanceVibe on self if aura is missing with 2000-6000 ms delay.

Thanks kw for the help here.


miraco:
For DanceVibe:
https://youtu.be/uHdpgzDF6Z0 

This wont fix the issue with combatlog getting spammed from this spell but thats not creature_ai related. It happens cause the spreading of the buff to allys in range gets called to often.

PR #1140, opened 2023-06-17
Update README.md

KurtRamsey:
Updated text, language, syntax, spacing, and fixed broken links. Working to review differences between classic/tbc/wotlk repositories to reconcile differences that are not expansion specific.


AnonXS:
Will be diffed with wotlk version.

AnonXS:
ok changed my mind.

PR #1156, opened 2023-08-11
[INSTANCE] shattered halls rework - part 1

miraco:
Add trash before last boss into spawn_group
remove assasin's pooling (will work with spawn_group when more spawns got added rework waypoints for assasins
rework Houndmaster + Warhound patrol movement (using spawn_group_formation)

using higher guids for now, because we start at the end of the instance, will reguid and cleanup when rework finished


AnonXS:
I would suggest breaking the rogue polling into more diff groups so they are never stacked

miraco:
There are more spawns between 1st and 2nd boss.
Only thing i know so far from all sniffs i had is that there are never more then 7 rogs spawned.
But on retail it was also possible that all 6 rogs before the last boss are up.

AnonXS:
The rogue patrols befor last boss are static

miraco:
> The rogue patrols befor last boss are static

Are you sure about that? 
This would mean there is only 1 rnd spawned assasin, cause in the way to endboss there are already 6.

Ill go and do some more sniffs to get more spawns/ see how its working on retail

PR #1175, opened 2023-12-15
fix(sql): Fixed use of aliases on `DELETE FROM` queries.

Byloth:
While running the [`InstallFullDB.sh`](https://github.com/cmangos/tbc-db/blob/master/InstallFullDB.sh) with option [`-World`](https://github.com/cmangos/tbc-db/blob/c9b50ab9c4b1f46a159b20a995647541eab00e71/InstallFullDB.sh#L2904-L2910), I experienced a syntax error on these 3 queries:

https://github.com/cmangos/tbc-db/blob/c9b50ab9c4b1f46a159b20a995647541eab00e71/Updates/0275_tbc_starter_zone_vein_overhaul.sql#L1041-L1043

The error output was:

```
ERROR 1064 (42000) at line 1041: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'a WHERE guid NOT IN(SELECT guid FROM gameobject b WHERE a.guid=b.guid) AND gu...' at line 1
```

---

This PR solve this syntax problem.

I'm using [MariaDB v10.10](https://hub.docker.com/_/mariadb) as database but this `ALIAS` syntax should work also on MySQL and even MSSQL databases ([source](https://stackoverflow.com/questions/10484532/delete-from-table-as-alias-where-alias-column-why-syntax-error#answer-26935749)).  
However, I haven't tested it personally on those databases.


evil-at-wow:
See also cmangos/issues#3619 for some discussion triggered by a similar issue in MySQL 5.7. I think the important point there is also the original query and how it doesn't look right at first sight (at least to me).

Nevertheless an interesting solution. I certainly didn't think about using multi-table deletes!

killerwife:
The bigger discussion is to be had if we still should support 5.7. But I digress, me being fancy bit me in the arse again.

evil-at-wow:
Well it's certainly a valid point because MySQL 5.7 has recently become EOL, so people should really move away from it if they haven't yet. But as long as the effort to keep it working is minimal I personally don't mind to make the occasional small change to fix things for them, but obviously any security implications are on them (even if the database is not publicly accessable).

The MariaDB 10.10 situation is similar by the way: maintenance has now stopped if I remember correctly, so time to migrate to MariaDB 10.11 :wink:

Byloth:
> [...]
> The MariaDB 10.10 situation is similar by the way: maintenance has now stopped if I remember correctly, so time to migrate to MariaDB 10.11 😉

Yeah... I know...
That's my bad...

But -I think- I'm gonna migrate directly to MariaDB 11.2.

celguar:
Did query standards also became EOL? WoW 2.4.3 is EOL for 15 years

celguar:
While all I know about mysql 8.0 is that it gets critical vulnerabilities fixed recently that did not exist in 5.7. MariaDB is a copy paste of mysql, I can't even use it portably, it requires me to change mysql.ini if I move the folder with mariadb...

PR #1182, opened 2023-12-27
[ACID][Instance] Shattered Hand Legionnaire - DeathScript rework

miraco:
This are the Database changes to https://github.com/cmangos/mangos-tbc/pull/648


ACID/acid_tbc.sql
```diff
@@ -24538,9 +24543,66 @@ INSERT INTO `creature_ai_scripts` (`id`,`creature_id`,`event_type`,`event_invers
 ('2092303','20923','4','0','50','0','0','0','0','0','0','0','54','0','0','11999','0','0','0','0','0','0','0','0','Blood Guard Porung - Random Say on Aggro'),
 ('2092304','20923','6','0','100','6','0','0','0','0','0','0','34','4','3','0','0','0','0','0','0','0','0','0','Blood Guard Porung - Finish Gauntlet on Death'),
 ('2092305','20923','7','0','100','6','0','0','0','0','0','0','34','4','2','0','0','0','0','0','0','0','0','0','Blood Guard Porung - Fail Gauntlet on Evade'),
-('2092306','20923','9','0','100','1029','0','5','7000','10000','0','0','11','15496','1','0','0','0','0','0','0','0','0','0','Blood Guard Porung - Cast Cleave (Heroic)');
-
-
+('2092306','20923','9','0','100','1029','0','5','7000','10000','0','0','11','15496','1','0','0','0','0','0','0','0','0','0','Blood Guard Porung - Cast Cleave (Heroic)'),
+-- guid based spawn_group_entry scripts
+-- Legionnaire 01
+('5400151','-5400151','6','0','100','2','0','0','0','0','0','0','53','5400030','0','0','0','0','0','0','0','0','0','0','spawn_group_entry (Normal) - Start RelayScript on Death'),
```
miraco, line 24549: have to call @AnonXS at this, this guids dont have a fixed entry so using entry*100 as id is kinda meh
but i also think guid*100 would result in to big numbers, i guess we need to find a good solution for this case

PR #1191, opened 2024-03-05
DM West: Illyanna Ravenoak movement.

evil-at-wow:
In Dire Maul West, Illyanna Ravenoak is currently a static spawn, but she should be walking around with her bear Ferra at her side (more or less). This PR adds her waypoints, with data coming from version 3.0.8 build 9506, and puts her in a formation with her bear.

Note: Ferra should be switching sides at the end of the path, but this currently doesn't happen. As far as I could see, it was not possible to do this when I added this, but I've recently seen some core change related to mirroring...

### DIFF
--- a/Updates/Instances/429_dire_maul.sql
+++ b/Updates/Instances/429_dire_maul.sql
@@ -1645,7 +1645,7 @@ INSERT INTO `creature` (`guid`, `id`, `map`, `position_x`, `position_y`, `positi
 (@CGUID+555, 11484, 429, -20.562, 700.975, -24.6149, 3.4988, 7200, 7200, 0, 0), -- Residual Monstrosity
 (@CGUID+556, 11486, 429, 132.626, 625.913, -48.3836, 4.62512, 7200, 7200, 0, 0), -- Prince Tortheldrin
 (@CGUID+557, 11487, 429, 33.14, 575.552, -4.31052, 4.53786, 7200, 7200, 0, 0), -- Magister Kalendris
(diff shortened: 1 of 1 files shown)


evil-at-wow:
My currently installed core doesn't support it so I can't test at the moment. But once I have updated my source tree again and have some time, I can check.

Grz3s:
mirroring was added in wotlk.

evil-at-wow:
It might indeed well be that the feature hasn't made it to the TBC core yet, I simply haven't checked to be honest.

Either way, I can look into it once it's available in all cores, but until then I would say this is a step forward even though it's not perfect yet.

Grz3s:
> the feature hasn't made it to the TBC core yet

and shouldnt be added. As I said.. mirroring was added in `wow 3.x.x` ..(wotlk era) .. 
Vanilla or TBC never had it... (or at least i never saw any) 

Side note:
Wotlk was confirmed recently.. in Icecrown. (thats why was added)

evil-at-wow:
That's an interesting remark. As I noted in my commit here, the path data here is also from the 3.x era. So that's not proving anything related to TBC or vanilla.

So I did some more research in the meantime, and it turns out that Ferra was always on the same side, being left, of Illyanna in TBC Classic (so wow 2.5) rather than switching sides at the end of her path. I think it's very reasonable to conclude that the switching sides is only needed in wotlk-db then.

al3xc1985:
also, check her aggro range. I tested her on wotlk, even thou I was 80 she still aggroes from like 30y like nohing

killerwife:
Mirroring is a db flag, so I think this is safe to merge @AnonXS i t hink.

PR #1194, opened 2024-03-21
Add playerbots db to the script

davidonete:


InstallFullDB.sh
```diff
@@ -209,8 +211,8 @@ function set_sql_queries
 
   # create database user and grant privileges
   SQL_CREATE_DATABASE_USER="CREATE USER IF NOT EXISTS '$MYSQL_USERNAME'@'$MYSQL_USERIP' IDENTIFIED BY '$MYSQL_PASSWORD';"
-  SQL_GRANT_TO_WORLD_DATABASE="GRANT SELECT, INSERT, UPDATE, DELETE, CREATE, DROP, ALTER, LOCK TABLES, CREATE TEMPORARY TABLES, EXECUTE, ALTER ROUTINE, CREATE ROUTINE ON \`$WORLD_DB_NAME\`.* TO '$MYSQL_USERNAME'@'$MYSQL_USERIP';"
-  SQL_GRANT_TO_CHAR_DATABASE=("GRANT SELECT, INSERT, UPDATE, DELETE, CREATE, DROP, ALTER, LOCK TABLES, CREATE TEMPORARY TABLES ON \`$CHAR_DB_NAME\`.* TO '$MYSQL_USERNAME'@'$MYSQL_USERIP';")
+  SQL_GRANT_TO_WORLD_DATABASE="GRANT INDEX, SELECT, INSERT, UPDATE, DELETE, CREATE, DROP, ALTER, LOCK TABLES, CREATE TEMPORARY TABLES, EXECUTE, ALTER ROUTINE, CREATE ROUTINE ON \`$WORLD_DB_NAME\`.* TO '$MYSQL_USERNAME'@'$MYSQL_USERIP';"
```
evil-at-wow, line 214: Out of curiosity: what is playerbots doing that it needs privileges to be able to create and drop indexes on the world and character databases?

There are other things that seem suspicious at first sight here - like the drop privilege - but that was already the case, so that's another discussion and not part of your changes.

davidonete: As far as I know we use indexes for random generation of bots and gear for them, to make it faster as there are thousands of randomisations happening every now and then. We could probably do it without it but we'll need to investigate.

evil-at-wow: I may be misunderstanding what these grant queries are doing. Making use of indexes is totally expected for performance reasons as you say. But there's a difference between having indexes available and setting them up or changing them from the core. 

My expectation is that user that sets up the databases ("the DBA", often 'root' for small setups in this project) runs the queries to set up the databases, tables, indexes etc. That's mainly what this install script does. The database user running queries in our software ('mangos', or whatever people choose) then merely makes use of the already set up indexes implicitly when running queries, but it doesn't need to set up new indexes or alter them. I understand form your answer this is indeed the case - generating new bots or gear for them is very likely limited to insert/update style queries.
 
The same goes for the 'drop' privilege I mentioned earlier: I doubt our core needs to be able to drop tables (or databases)! And maybe some others. But none of that is related to this change of course.

Now, as I mentioned, I may be misunderstanding things at the moment, so I'd have to check more carefully. All I wanted to say is that the running core - or more specifically the database user it's using to connect to the database - should have the minimal privileges to do its job, and nothing more.

PR #1211, opened 2024-06-23
[ACID] BF: rewrite The Maker into acid and spell_list

miraco:
add db changes for this change:
https://github.com/cmangos/mangos-tbc/pull/681

### DIFF
--- a/Updates/04345_boss_the_maker.sql
+++ b/Updates/04345_boss_the_maker.sql
@@ -1,0 +1,19 @@
+-- SpellList for The Blood Furnace - The Maker
+-- Some Infos:
+-- Domination can also hit Tanks
+DELETE FROM `creature_template_spells` WHERE `entry` IN (17381, 18621);
+
+DELETE FROM `creature_spell_list_entry` WHERE `Id` IN (1738101, 1862101);
+INSERT INTO `creature_spell_list_entry` (`Id`, `Name`, `ChanceSupportAction`, `ChanceRangedAttack`) VALUES
+(1738101, 'Blood Furnance - The Maker', 0, 0),
+(1862101, 'Blood Furnance - The Maker', 0, 0);
+
+DELETE FROM `creature_spell_list` WHERE `Id` IN (1738101, 1862101);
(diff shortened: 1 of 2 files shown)


Updates/04345_boss_the_maker.sql
```diff
@@ -0,0 +1,19 @@
...
+DELETE FROM `creature_template_spells` WHERE `entry` IN (17381, 18621);
+
+DELETE FROM `creature_spell_list_entry` WHERE `Id` IN (1738101, 1862101);
+INSERT INTO `creature_spell_list_entry` (`Id`, `Name`, `ChanceSupportAction`, `ChanceRangedAttack`) VALUES
+(1738101, 'Blood Furnance - The Maker', 0, 0),
+(1862101, 'Blood Furnance - The Maker', 0, 0);
+
+DELETE FROM `creature_spell_list` WHERE `Id` IN (1738101, 1862101);
+INSERT INTO `creature_spell_list` (`Id`, `Position`, `SpellId`, `Flags`, `CombatCondition`, `TargetId`, `ScriptId`, `Availability`, `Probability`, `InitialMin`, `InitialMax`, `RepeatMin`, `RepeatMax`, `Comments`) VALUES
+(1738101, 1, 30923, 0, -1, 100, 0, 100, 0, 15000, 26000, 20000, 30000, 'The Maker - Domination - random'),
```
killerwife, line 13: Pretty sure you introduced a bug here, target 100 takes random target in aggro, but core has ATTACKING_TARGET_RANDOM, 1, nullptr, SELECT_FLAG_PLAYER - which is random target except topaggro (the 1 means skip 1 from top) which is target 101 if i remember correctly.

killerwife: 101	1	0	1	2	-1	Attack - random player non tank 

100	1	0	0	2	-1	Attack - random player

notice the 1 diff :D

miraco: Yes and no, we had it on non tank before, but testing it on wotlk ptr showed that it can also hit tanks.

PR #1217, opened 2024-07-12
Fix 'The Lady Necklace' item not dropping

miraco:
As far as i understand our loot_template, negative ChanceOrQuestChance means Player needs to have Quest so item can drop, change it to normal and slightly increase/decrease Chance.

@AnonXS you are our loot pro, pls confirm that im right with this.


AnonXS:
yea fel into the trap and just increased chance, not thinking about if its actually lootable.

grafik

PR #1248, opened 2025-03-18
Git: Update Cache Module to v4, Update MySQL to 8.0.41 (LTS)

insunaa:
tryfix release build of database


insunaa:
The MySQL update is optional, tho we should at least change the version to 8.0.41

PR #1266, opened 2025-10-02
npc_text_broadcast_text

lanfire001:
Add missing lines to table npc_text_broadcast_text to show translated gossip text translations instead of original english text.


lanfire001:
Needs to be double checked before merge.

PR #1288, opened 2026-05-13
Drop stale locales_gameobject structure dump

mserajnik:
The mysqldump output re-created `locales_gameobject` with its pre-s2485 columns (`castbarcaption_loc1..8`), reverting the rename done by [`mangos-tbc:sql/updates/mangos/s2485_01_mangos_closing_text.sql`](https://github.com/cmangos/mangos-tbc/blob/a16a48b1e37db278f714d8cb009031f359663c2e/sql/updates/mangos/s2485_01_mangos_closing_text.sql) and breaking fresh installations (`mangosd` aborts on `opening_text_loc1`).

This also switches the `INSERT`s to an explicit column list so future column additions like `closing_text_loc*` keep working without re-touching this file.


killerwife:
Just fyi, for this to work properly, you need to edit the file to be fully compliant with target result. So fully incorporating the new structure to the file.

This is due to the order of operations in install script:

image

Updates are done BEFORE locales, so locales are expected to be in already correct final format as core expects them. So unless im horribly mistaken, the revert of the naming will still break core :)

mserajnik:
Thanks for taking a look!

Maybe I misunderstood something, but this was my interpretation and is what the PR is based on:

The order you describe is exactly what makes the PR work: by the time `apply_locales` runs, `locales_gameobject` is already in its post-s2485 shape, set up by [`Updates/0698_cls_rework.sql`](https://github.com/cmangos/tbc-db/blob/39bf73590beb0e878364e952353b632ca630bc27/Updates/0698_cls_rework.sql) running earlier in `apply_content_db`. Dropping the `DROP` + `CREATE` here lets the locales step inherit that schema instead of reverting it.

The `INSERT` statements are also rewritten to use an explicit column list mapping the positional values to the new column names:

```sql
INSERT INTO `locales_gameobject` (
  `entry`,
  `name_loc1`, ..., `name_loc8`,
  `opening_text_loc1`, ..., `opening_text_loc8`
) VALUES (...);
```

With this PR, there is no rename; `castbarcaption_loc*` is gone from the file entirely.

Happy to also update the file to fully re-include a `CREATE TABLE` block (matching the new schema) if you'd prefer that style over relying on the upstream migrations but I went with the explicit `INSERT` option because it keeps the locales file robust against future column additions without further edits.

killerwife:
I misunderstood then cos github sucks at large diffs. Over weekend ill try to test and merge accordingly

PR #1290, opened 2026-05-31
Create 0722_ring_of_blood_quest.sql

Velite12:
Fix ring of blood quest flag


AnonXS:
Quests have Flag=131 in ccsdb_tbc. confirmed with tbc sniff data, this fix is incorrect, will use custom solution instead.

Velite12:
Yes I made a custom solution that works, but the flag has to be removed. Check https://github.com/cmangos/mangos-tbc/pull/817 where I had to make the pr after the merge.

Velite12:
In future, maybe in core, check if quest is a group quest, so if it has flag for staying alive, that needs to be interpreted as "group needs to be alive". The way I implemented it, is if the mob resets/despawns without kill, it's a fail.

killerwife:
What anon means is that sniffs show data to have that flag. Needs more research on the intricacies for the flag. For now good enough.

PR #1292, opened 2026-06-08
Fix imps being unable to autocast blood pact while in phase shift

Velite12:
Imps can cast blood pact while in phase shift, but they are unable to autocast due to autocast checking immunity in the core. Since an imp will first phase shift before blood pact, it sees that it's immune. However it is possible to manually click on blood pact to cast it _while in phase shift_.

This probably needs an extensive look and cleanup. In the meantime, this allows imps to autocast blood pact when summoned if phase shift is also set to autocast.


killerwife:
Nope. Exactly what you wrote in comment has to be done. We dont alter spells to avoid the dirty work :D

PR #1293, opened 2026-06-28
Cipher of Damnation final quest fixes

Velite12:
Fix DB issues with Cipher of Damnation quest 10588. This is a companion PR with Core changes: https://github.com/cmangos/mangos-tbc/pull/834

Video of quest:
https://www.youtube.com/watch?v=e5kLAoug2Ms
https://youtu.be/4wwSlUF7QV8
https://youtu.be/4wwSlUF7QV8

Summary:

https://www.wowhead.com/tbc/npc=21685/oronok-torn-heart 

move oronok torn-heart abilities to db with targeting for healing wave for friendlies

fix incorrect stat values for oronok and borak, gromtor hp was correct
https://www.wowhead.com/tbc/npc=21686/borak-son-of-oronok
https://www.wowhead.com/tbc/npc=21685/oronok-torn-heart

create guids for oronok and sons

add spawn group and formation for oronok and sons. Creature ids 21685 21686 21687

They are supposed to move in Geese type formation, and the current following causes the 2 npcs to be stuck when they fly in, which causes them to not path correctly in the beginning.

remove old creature_movement_template path because we use waypoint_path for the spawn group

spawn the spawn group on event (using the item)

### DIFF
--- a/Updates/0735_cipher_of_damnation_fixes.sql
+++ b/Updates/0735_cipher_of_damnation_fixes.sql
@@ -1,0 +1,93 @@
+# fix issues with Cipher of Damnation final quest
+
+# https://www.wowhead.com/tbc/npc=21685/oronok-torn-heart move oronok torn-heart abilities to db with targeting for healing wave for friendlies
+
+DELETE FROM creature_spell_list WHERE Id = 2168501;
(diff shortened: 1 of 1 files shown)


Velite12:
Updated to be compliant, let me know if anything else needs to be modified. Quest is currently bugged without these changes as core part is merged already.

AnonXS:
Core: https://github.com/cmangos/mangos-tbc/commit/9c273b3684981db3f35e95ea830105b3bc69928b

DB: https://github.com/cmangos/tbc-db/commit/da2de07e6606d495872c3fd92ba8363cf79f43c9

Seen spells like 37246    Oronok Speed Increase being used, but that can be done another time.

PR #1294, opened 2026-07-01
[Instance] Fix left group at Mechanar entrance missing flag

Velite12:
Fix left group at entrance missing aggro together flag


miraco:
Can you rename the pr to something like

[Instance] Fix Mechanar entrance group flag

In general if we change a file inside instance folder wie use [Instance] for acid changes we use [ACID] 

else pr looks good

PR #1296, opened 2026-07-13
Fix underbog Bog Giant and Underbog Lord wrong spells

Velite12:
Remove Trample (spell 2) from Heroic Bog Giant, move spell 3 to spell 2 in its place.

Add Enrage spell at low hp (seems below 30) to Normal Bog Giant.

Proof:
Normal Underbog Bog Giant:
https://youtu.be/_LRpztnnao8?t=241

You can see that it does the trample, all of the melee are affected (they have like 3 red spikes above their head), the tank dodges the attack. There is a yellow glow on the floor that comes out from the mob, and the screen shakes to show the effect. There's also a sound effect that goes with it too. This is also how the spell is shown in cmangos.

You can also see that when it gets low hp (around 25 but it jumps quickly), it has an Enrage spell.

Heroic Underbog Bog Giant:
https://youtu.be/kvXZdn8ePTo?t=897

You can see that during the course of killing this mob, it never does this Trample spell (there is no screen shake, sound, or effect above head of characters to show the effect), nor damage dealt (it currently deals 5k+ damage aoe in Heroic and can scale quickly due to the Growth spell damage modifier, understandably this is probably why it is not present in Heroic).

You can also see that there is a difference in the spells listed on wowhead from normal to heroic:

https://www.wowhead.com/tbc/npc=17723/bog-giant#abilities;mode:normal
https://www.wowhead.com/tbc/npc=17723/bog-giant#abilities;mode:heroic

Underbog Lord also has this same enrage spell, but at a lower HP, in Normal Mode:

https://youtu.be/_LRpztnnao8?t=1893
https://www.wowhead.com/tbc/npc=17734/underbog-lord#abilities;mode:normal

### DIFF
--- a/Updates/0738_underbog_fixes.sql
+++ b/Updates/0738_underbog_fixes.sql
@@ -2,9 +2,3 @@
 -- Remove Trample (spell 2) from Heroic Bog Giant, move spell 3 to spell 2
 DELETE FROM `creature_spell_list` WHERE `Id` = 2016401 AND `Position` = 2;
 UPDATE `creature_spell_list` SET `Position` = 2 WHERE `Id` = 2016401 AND `Position` = 3;
-
--- Add Enrage at low hp (seems below 30) to Normal Bog Giant
-INSERT INTO `creature_spell_list` (`Id`, `Position`, `SpellId`, `Flags`, `CombatCondition`, `TargetId`, `ScriptId`, `Availability`, `Probability`, `InitialMin`, `InitialMax`, `RepeatMin`, `RepeatMax`, `Comments`) VALUES ('1772301', '3', '8599', '0', '5364', '0', '0', '100', '1', '0', '0', '0', '0', 'Bog Giant - Enrage below 30');
-
--- Add Enrage at low hp (seems below 15) to Normal Underbog Lord
-INSERT INTO `creature_spell_list` (`Id`, `Position`, `SpellId`, `Flags`, `CombatCondition`, `TargetId`, `ScriptId`, `Availability`, `Probability`, `InitialMin`, `InitialMax`, `RepeatMin`, `RepeatMax`, `Comments`) VALUES ('1773401', '3', '8599', '0', '1107', '0', '0', '100', '1', '0', '0', '0', '0', 'Underbog Lord - Enrage below 15');
--- a/Updates/0738_underbog_fixes.sql
+++ b/Updates/0738_underbog_fixes.sql
@@ -5,3 +5,6 @@ UPDATE `creature_spell_list` SET `Position` = 2 WHERE `Id` = 2016401 AND `Positi
 
 -- Add Enrage at low hp (seems below 30) to Normal Bog Giant
 INSERT INTO `creature_spell_list` (`Id`, `Position`, `SpellId`, `Flags`, `CombatCondition`, `TargetId`, `ScriptId`, `Availability`, `Probability`, `InitialMin`, `InitialMax`, `RepeatMin`, `RepeatMax`, `Comments`) VALUES ('1772301', '3', '8599', '0', '5364', '0', '0', '100', '1', '0', '0', '600000', '600000', 'Bog Giant - Enrage below 30');
+
+-- Add Enrage at low hp (seems below 15) to Normal Underbog Lord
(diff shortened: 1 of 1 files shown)


killerwife:
I did a quick check rn of underbog bog giant on hc. I do not see any Double Attack, Trample or other ability, just like you said. However gonna leave this merge to miraco just in case i missed smth

Velite12:
Well I'll say this, with the ability damage calculation and the growth damage buff, the way you would notice it in classic is tanks getting near 1 shot on the 2 pack right before Hungarfen. You have to tank two of them at once. It's not as bad in normal because it's a low HP enrage and not a damage buff throughout the fight.

It's a diceroll if you dodge/parry the ability or not currently, otherwise you just have to burn cds on one, wipe, run back and kill the other. Doable but definitely more difficult.

killerwife:
There is one issue however, classic tbc has hcs in prenerf state rn. So the info has to be viewed with a bit of scepticism. Default tuning in tbc-db is postnerf.

Velite12:
If "post nerf" is harder than "pre nerf"... where's the nerf. But I will look for original tbc too.

Velite12:
https://www.youtube.com/watch?v=4LMfs_Okc3g
image


I didn't manage to get around to this one, but it would appear that in Original TBC, these underbats are doing less damage than the ones we have currently. Seems like 3.3k on a naked character in heroic, but I have them around 5-7k on a naked character in heroic in ours:

image

https://youtu.be/kvXZdn8ePTo?t=372

t4 warrior tank we're talking mid to upper 1000s autos.

vs on blue prot paladin in cmangos:

image

So something to look at.

These are all on latest commit. I see they have a damage multi of 21.5345.

ordinarily not a big deal, buuuuuuuuuut they do have an ability that does 400% weapon damage... which results in about 11-12k damage on same paladin. Condition is if they are behind you but I haven't seen anybody taking that much damage in classic.

So the autos being below 2k is the difference of almost 4-5k damage on the ability, therefore out of 1 shot range.

image
If we take the damage multiplier down to 15.5... (2x of normal mode creature damage multplier) we start to get nearer to this tbc video. But this is just my conjecture.

Velite12:
https://youtu.be/CRMNLQtsxig?list=RDCRMNLQtsxig&t=72

best proof here: heroic underbog in Mar 26, 2007, there's two rogues and they take no large hits that would follow from Trample damage in heroic. No dodges or parry or miss text appears by their unit frames either. I think towards the end they take some very small damage from something else but definitely not the trample, not like in our version (2x trample same time kills me, a rogue). So it never really existed to begin with, I think.

miraco:
When i reworked those dungeons from creature_ai into spell_list i most of tihe time just converted it - so yeah its possible that those npcs dont use it on heroic.
You can see in the commit, before they had the trample for both in heroic and normal already in creature_ai_script

however both npcs should still have the enrage spell
`('1772301','17723','2','0','100','1027','30','0','120000','120000','0','0','11','8599','0','0','1','2384','0','0','0','0','0','0','Bog Giant (Normal) - Cast Enrage at 30% HP'),
('1773403','17734','2','0','100','1027','30','0','120000','120000','0','0','11','8599','0','0','1','2384','0','0','0','0','0','0','Underbog Lord (Normal) - Cast Enrage at 30% HP'),`

Both done via creature_ai_scripts - mostly cause they should only cast this once, so your way with adding it spell_list with a 600000 timer is wrong cause they would use it again after this time

I am gonna look at this when im done with my heroic research on tbc anni

btw: only use [instance] as tag if you change something inside tbc-db/updates/instance folder pls.

Velite12:
I ripped the example from orgrimmar grunt who uses it with similar conditions and same spell. But no, these were not present in the db.
image

1773401 is the creature spell list for underbog lord, not 1773403. There is nothing for 1773403 in creature_spell_lists.

image

Can see the 3 entries for the 2 spells + the one I added back in. But they don't do the enrage at 30% hp like the bog giant.

miraco:
Because its creature_ai_scripts not creature_spell_list for enrage

Velite12:
I see, so then we just need to remove trample.

AnonXS:
https://github.com/cmangos/tbc-db/commit/0e5f241bbdd1439b42a0c81282e46405f4eba6c8

PR #1299, opened 2026-07-29
Update dmg values to our current calculation based on bestiary values

miraco:
Calculation based with our current known values, will update it with Karazhan

This is open for discussion for now

### DIFF
--- a/Updates/0739_t4_dmgvalues.sql
+++ b/Updates/0739_t4_dmgvalues.sql
@@ -1,0 +1,132 @@
+-- Update DamageMultiplier and DamageVariance to match with newest core changes
+-- Damage Values taken from WoW Bestiary
+-- Also adding the outcome (on 0 armor) with our current calculation after rounding
+
+-- ------------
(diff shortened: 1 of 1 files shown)


AnonXS:
Still interesting to see what it was befor, /*so update the other column so we have a hard compare even preserver what that was befor that*/, if that last compare is not even included. we need comparrison.

miraco:
Well, we now have 3 different sets of values.
Min/Max DMG, which were hardcoded in the DB and from which the damage multiplier was initially calculated.
Then, from the old formula, DamageMultiplierOLD and DamageVarianceOLD, which also used BaseDamageExp1OLD.
And the new DamageMultiplier and DamageVariance.
Which of the 3 datasets do you want? And what do you want to compare it with? I don't really see the point in documenting the old, incorrect values, since we know pretty much exactly what the outcome should be — which is also documented (calculated min/max).
I deliberately didn't add the old stuff, because it doesn't personally interest me and I only work with the things we know.

PR #1300, opened 2026-08-05
Chess event fixes

Velite12:
Companion PR to https://github.com/cmangos/mangos-tbc/pull/848

Address orc chess spells targeting wrong units ~and address the chess units needing to have 0 armor (therefore 0 armor multiplier and 0 agility multiplier).~ Chess pieces confirmed to have armor by sniff so the problem of their spells being reduced by armor has to be handled elsewhere.


miraco:
0 Armor is defnitly wrong here
Confirmed from TBC Anni today
WowClassic_lgQc6cKUpn

only made it to take 2 pieces but both had armor.

Velite12:
So they have armor, but the armor does nothing? How do we do that?

Do we have to create some chess specific exception to armor calculation? Or spell specific? Is there even a flag to do this with, since the spells are considered physical school. Do they have armor penetration?

The point is, we know the outcome, we know the inputs now, but we don't know, and possibly can't know, the way to go from inputs -> outcome, especially if this happens to be script specific. Unless there is more evidence, any solution to the problem we have to address is going to have to be guessed.

Obviously you would have seen that the physical abilities are not mitigated by armor.

Velite12:
What is their agility then? Since the agility multiplier affects the creature's armor but we have to ignore that too it seems.

